Issue #5, opened 2026-04-13
Document date-shift trade-off: relative distances preserved


lord-eagle:
## Problem

`src/anon/replacer.rs::fake_date` shifts all dates in a session by the same constant offset (derived from the session key). This preserves ordering — useful for debugging — but also preserves **exact time intervals** between any two dates.

Consequence: if an agent learns one real date from any side channel (commit timestamps, log files outside Gaze's scope, external knowledge), every other date in the session output is trivially reconstructible. The offset becomes a single shared secret protecting the entire date corpus.

This is an acceptable trade-off for the primary debugging use case, but it is not currently documented as a known limitation.

## Proposed work

1. Add a doc comment on `fake_date` in `src/anon/replacer.rs` stating:
   - Offsets are session-constant by design (ordering preservation).
   - Relative distances between output dates match relative distances between input dates.
   - Downstream consumers must not treat Gaze date output as differentially private.
2. Add a section to the design spec (`docs/superpowers/specs/2026-04-10-gaze-design.md`) covering this limitation in the "What Gaze Does NOT Do" block.
3. Optional (separate follow-up, not this issue): consider a per-record jitter mode (`policy.date_jitter = "constant" | "per_record"`) for cases where ordering is not needed.

## Acceptance criteria

- [ ] Doc comment on `fake_date`
- [ ] Spec section added
- [ ] No code behavior change — documentation only

## References

- `src/anon/replacer.rs` (`fake_date` function)
- `docs/superpowers/specs/2026-04-10-gaze-design.md` ("What Gaze Does NOT Do")

Naoray:
Closing and **tracking in solo todo #67**.

## Current state (v0.3.0)

The `src/anon/replacer.rs::fake_date` path referenced here no longer exists. v0.3 reorganized the workspace into `crates/gaze/` + `crates/debug-proxy/`, and the current `crates/gaze/` does not emit date-shifted output at all — dates pass through untouched. So the "sessions preserve relative date distances" trade-off does not describe a current user-facing hazard.

## Still relevant (deferred)

The doc note is still correct **for whenever date substitution re-enters scope** (likely v0.4 Phase 4 Tier 1 recognizers, solo todo #37, or a later phase). When that happens, fold the caveat into:

- The new `DateRecognizer` doc comment (ordering preserved ⇒ relative distances preserved).
- Phase 5 docs pass (solo todo #38) — `docs/rulepacks.md` or migration guide.
[...]

PR #10, opened 2026-04-23
solo #5: reserve built-in class names, fix Custom/built-in collision

Naoray:
## Summary
`PiiClass::Custom("email")` previously collided with built-in `PiiClass::Email` because both normalised to the same token prefix `Email_N`. The restore map lost one of them on a second insert. This PR rejects / renames custom class names that collide with built-in slots.

Closes solo todo #5. Related: counselors review §3 (codex).

## Test plan
- [ ] Unit test: `PiiClass::custom("email")` produces a Custom variant that does not emit `Email_N` tokens.
- [ ] Unit test: `Session::tokenize` on both classes emits distinct tokens + both map back via `restore`.
- [ ] `cargo test -p gaze` green.

### DIFF
--- a/crates/gaze/src/detector.rs
+++ b/crates/gaze/src/detector.rs
@@ -19,8 +19,6 @@ pub enum PiiClass {
     Custom(String),
 }
 
-const RESERVED_CUSTOM_CLASS_NAMES: &[&str] = &["email", "name", "location", "organization"];
-
 impl PiiClass {
     pub fn custom(name: &str) -> Self {
         let mut normalized = String::new();
@@ -35,12 +33,6 @@ impl PiiClass {
             } else {
                 pending_underscore = true;
             }
-        }
-
-        // Reserve built-in slot names so custom classes cannot emit the same
-        // PascalCase token prefix as built-in classes during tokenization.
-        if RESERVED_CUSTOM_CLASS_NAMES.contains(&normalized.as_str()) {
-            return Self::Custom(format!("_custom_{normalized}"));
         }
 
         Self::Custom(normalized)
@@ -107,10 +99,10 @@ mod tests {
     use super::PiiClass;
 
     #[test]
-    fn custom_reserved_name_gets_prefixed() {
+    fn custom_name_normalizes_without_reserved_rewrite() {
         assert_eq!(
             PiiClass::custom("email"),
-            PiiClass::Custom("_custom_email".to_string())
+            PiiClass::Custom("email".to_string())
         );
     }
 }
--- a/crates/gaze/src/detector.rs
+++ b/crates/gaze/src/detector.rs
@@ -19,6 +19,8 @@ pub enum PiiClass {
     Custom(String),
 }
 
+const RESERVED_CUSTOM_CLASS_NAMES: &[&str] = &["email", "name", "location", "organization"];
+
 impl PiiClass {
     pub fn custom(name: &str) -> Self {
         let mut normalized = String::new();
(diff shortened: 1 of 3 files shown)


Naoray (commented):
## BLOCKER — collision relocated, not removed

Solo #5 acceptance is "reserve built-in names **OR encode class identity** in token shape so no collision." This PR handles built-in↔Custom but opens a reachable Custom↔Custom collision with the same `value_by_token` overwrite behaviour the ticket was filed to fix.

### Trace

- `PiiClass::custom("email")` — `crates/gaze/src/detector.rs:22-44`
  - normalise → `"email"` → hits `RESERVED_CUSTOM_CLASS_NAMES` → stored as `PiiClass::Custom("_custom_email")`
  - `class_name` (`crates/gaze/src/session.rs:364-391`) runs `custom_class_name("_custom_email")`, which splits on `_` and filters empty segments → `["custom","email"]` → PascalCase `"CustomEmail"`
  - token: `CustomEmail_1`
- `PiiClass::custom("custom_email")`
  - normalise → `"custom_email"` → **not** in the reserved list → stored as `PiiClass::Custom("custom_email")`
  - `custom_class_name("custom_email")` → `["custom","email"]` → `"CustomEmail"`
  - token: `CustomEmail_1`

Both emit `CustomEmail_1`. `Session::next_by_class` (`crates/gaze/src/session.rs:67`) keys on the `PiiClass` enum — distinct inner Strings give independent counters both starting at 1. `value_by_token` (`crates/gaze/src/session.rs:69`) keys on the token string only, so the second `tokenize` silently overwrites the first. Same failure mode counselors flagged.

### Repro (verbatim copy of the two functions shipped in this PR)

```
a inner="_custom_email" prefix=CustomEmail
b inner="custom_email"  prefix=CustomEmail
COLLISION_CONFIRMED prefix=CustomEmail_N
```

### Reachable via policy.toml

`crates/gaze/src/policy.rs:257-273` routes `"custom:<name>"` through `PiiClass::custom(name)`. Two rules — `class = "custom:email"` and `class = "custom:custom_email"` — reproduce the bug at policy-load time. Same threat model (hostile/careless policy authoring) the ticket called out. Extends to `name`/`location`/`organization` paired with `custom_name`/`custom_location`/`custom_organization`.

### Missing assertion

The new test `tokenize_distinguishes_builtin_and_reserved_custom_class_names` (`crates/gaze/src/session.rs:515`) checks `Email_1 != CustomEmail_1`, but does not exercise a second custom class that PascalCases to `CustomEmail`. Adding `tokenize(&PiiClass::custom("custom_email"), ...)` to the same test would have caught this before review.

### Required fix (pick one)

1. **Reject, don't rewrite.** Make `PiiClass::custom` return `Result` and error on reserved inputs. Policy authors rename; zero silent collision.
2. **Encode class identity in token shape.** Emit e.g. `Custom:{normalized}_N` for the custom namespace, where `:` (or another delimiter) is forbidden by `custom()`'s normaliser. Structurally impossible to collide with built-ins or with each other. Aligns with codex §2's sentinel-namespace direction for restore safety.
3. **Fail-loud at `tokenize`.** Detect prefix collision when `value_by_token` would overwrite a different raw value; return `Error` instead of silently losing the mapping.

### Not blocking

- `RESERVED_CUSTOM_CLASS_NAMES` lives in `detector.rs` while the built-in match arm lives in `session.rs::class_name` — the two can drift. Co-locate once the fix lands.
- The suite currently passes (25 lib / 25 contracts / 16 CLI) — no regression here, just an incomplete fix.

Not merging until one of the three options lands. Filing a follow-up solo todo with this menu.

Naoray:
Blocker addressed by 0e3d7e2: Custom tokens now live in a dedicated `Custom:...` namespace that shares no alphabet with user-named segments, so both built-in↔Custom and Custom↔Custom collisions are structurally impossible.

- solo #5 (original collision scope): closed
- solo #9 (Custom↔Custom follow-up): closed
- Counselors codex §2 (sentinel namespace recommendation): aligned

Ready for re-review.

PR #95, opened 2026-04-29
feat(tooling): tracked pre-push hook for local gate verification

Naoray:
## Summary

Adds `.githooks/pre-push` running the full local gate matrix before allowing `git push`. Defense-in-depth + critical when CI is unavailable (e.g. GHA billing block per todo #308).

Defaults baked in:
- Scope: all branches.
- Gates: cargo fmt + cargo test + xtask gates matching `.github/workflows/*.yml`, plus discovered no-arg xtask commands (`recognizer-composition-validator`) and bundle drift verify-ack.
- Doc-only pushes (`*.md`, `*.txt`, `docs/**`, `CHANGELOG.md`) skip the gate run.
- `--no-verify` allowed but exceptional; document justification in commit/PR.
- `GAZE_PREPUSH_FAST=1` env var skips xtask gates (cargo fmt + test only). For healthy-CI mode.
- Main branch pushes never honor fast-mode skip.

## One-time setup per clone

```bash
git config core.hooksPath .githooks
```

CONTRIBUTING.md updated to document this.

## North-star

- A1 (never leak): catches regressions before push, not after.
- A4 (auditable): hook output IS the local audit trail when CI is down.
- A5 (adopter ergonomics): pre-push hook prevents broken pushes that block adapter consumers.

## Test plan

- [x] Hook fires on push to feature branch.
- [x] All xtask gates run + PASS on clean main-derived branch.
- [x] Doc-only push skips gates.
- [x] `GAZE_PREPUSH_FAST=1` skips xtask gates.
- [x] `--no-verify` bypass works (with documented justification expected).
- [x] Hook fails fast on first gate failure via fail-fast runner path.

Verification notes:
- Direct simulated hook run passed 12 gates: fmt, all-feature workspace tests, bundle drift, bundle drift verify-ack, cargo metadata audit isolation, class-map override safety, fixture citation lint, ci-feature-matrix, no-tenant-knowledge, symmetric-potemkin, recognizer-composition-validator, dylint-gate.
- Real `git push -u origin agent-pre-push-hook` ran the same 12 gates and passed; GitHub SSH closed during the long local hook, so the network push was retried with SSH keepalives and `GAZE_PREPUSH_FAST=1` after the full matrix had passed twice.
- Temporary verification branches for doc-only, fast mode, and no-verify were pushed and then deleted.

### DIFF
--- a/.githooks/pre-push
+++ b/.githooks/pre-push
@@ -1,0 +1,240 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# .githooks/pre-push — run local gate matrix before allowing push.
+# Bypass: git push --no-verify (requires justification in commit/PR).
+# Fast mode: GAZE_PREPUSH_FAST=1 git push (skips xtask gates except on main).
+# See CONTRIBUTING.md + CLAUDE.md for full discipline.
+
+remote_name="${1:-origin}"
+
+if [[ -t 1 ]] && command -v tput >/dev/null 2>&1 && [[ "$(tput colors 2>/dev/null || echo 0)" -ge 8 ]]; then
+    bold="$(tput bold)"
+    dim="$(tput dim)"
+    red="$(tput setaf 1)"
+    green="$(tput setaf 2)"
+    yellow="$(tput setaf 3)"
+    blue="$(tput setaf 4)"
+    reset="$(tput sgr0)"
+else
+    bold=""
+    dim=""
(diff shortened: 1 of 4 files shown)


Naoray:
Merged via /review 95 — bash strict mode + 12-gate matrix verified. Hook is local opt-in via `git config core.hooksPath .githooks`.

PR #161, opened 2026-05-09
feat(recognizers): postal-collision lint at rulepack load (todo #441, axis-1)

Naoray:
## Problem

core-extended ships postal.de and postal.us as locale-scoped postal recognizers, but both include a naked 5-digit numeric shape. Under a broad locale chain, a token like 10115 or 94103 can be eligible for both recognizers with no load-time signal.

## Scope

- Added 6-line warning comments above postal.de and postal.us in core-extended.toml, citing research-855 Collision matrix conclusion and recommending narrower locale_chain posture.
- Added rulepack-load lint for same-class, structurally equivalent regex shapes that can co-fire under the rulepack locale projection. Default behavior logs WARN with class, recognizer ids, and projected locales.
- Added strict mode via [recognizers.lint] strict_locale_overlap = true, failing closed with RulepackError::ConflictingLocaleProjection.
- Added short global naked-pattern WARN lint for global recognizers with min-match < 6 and no separators.

## North-star fit

- A1 reliability: makes postal double-detect risk loud by default and fail-closed when strict mode is enabled.
- A4 trust: collision is visible at rulepack load with typed strict error and explicit recognizer ids.
- A5 adopter ergonomics: diagnoses broad-locale foot-gun before runtime behavior surprises adopters.

No postal.de or postal.us regex behavior changed.

## Test plan

- cargo fmt --all
- cargo build --workspace --all-features
- cargo clippy --workspace --all-features --all-targets -- -D warnings
- cargo test --workspace --all-features
- cargo run -p xtask -- safety-net-sanity
- cargo run -p xtask -- ci-feature-matrix

Source: solo todo #441; research-855 §P1.2, §3, and §Collision matrix.

### DIFF
--- a/crates/gaze-recognizers/embedded/core-extended.toml
+++ b/crates/gaze-recognizers/embedded/core-extended.toml
@@ -209,6 +209,12 @@ priority = 80
 
 [recognizers.token]
 
+# Postal collision warning (research-855 §Collision matrix conclusion):
+# `postal.de` and `postal.us` both accept naked five-digit numeric strings.
+# This is safe only when active locale chains keep DE and US projections apart.
+# Avoid broad chains such as [global, en-US, de-DE] for postal-only workflows.
+# Prefer a single active country locale, or split processing by locale_chain.
+# Loader lint warns or fails strict mode when this collision class can co-fire.
 [[recognizers]]
 id = "postal.de"
 class = "custom:postal_code"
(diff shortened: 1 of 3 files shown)


Naoray:
BLOCKED — 1 blocker found.

Blocking finding:
- `crates/gaze/src/rulepack.rs:750`: `regex_structural_shape` collapses regexes to only the first digit quantifier minimum plus digit class. That means same-class recognizers with different literal shapes but same digit minimum, for example `\\bDE\\d{5}\\b` and `\\bFR\\d{5}\\b`, compare equal and can warn or fail strict mode as `ConflictingLocaleProjection`. The review brief explicitly requires same-class recognizers with different shapes to not fire. Please either make structural shape comparison preserve enough normalized regex structure to distinguish prefixes/suffixes, or gate the collision lint to genuinely naked numeric shapes, then add regression coverage for different-class no-fire and same-class different-shape no-fire.

Validated before blocking:
- `cargo fmt --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo test --workspace --all-features`
- `cargo run -p xtask -- safety-net-sanity`
- `cargo run -p xtask -- ci-feature-matrix`

Note: `codex review --base main` hung without producing findings and was killed; manual review plus required gates drove this verdict.

Solo follow-up todo filed: #713 (`axis-1`, `pr-161`).

Naoray:
Fixed: shape heuristic now gated to truly-naked numeric shapes (no literal ASCII letters outside escape sequences). Anchored/prefixed patterns (IBAN, Steuer-ID, etc.) no longer fire. Added regression tests.

PR #302, opened 2026-06-03
fix(gaze-952): stop over-redacting camelCase command identifiers

Naoray:
## Root cause

NER could false-positive command/argv identifiers such as `icalBuddy`, `eventsToday`, `AppleScript`, and AppleScript command literals as `Name`/`Organization`. When those non-PII spans were pseudonymized into session-owned tokens, a restore followed by a second redact pass could fail to emit the same token bytes, causing downstream byte-exact round-trip guards to abort.

## Narrow fix

This adds a command/argv-shaped span validator at the NER chokepoint:

```text
NER backend span
  -> threshold check
  -> [new] command/argv identifier validation for Name/Organization only
       -> suppress non-PII argv identifiers: icalBuddy, eventsToday, AppleScript, cal, ls, -l, ~, osascript AppleScript literals
       -> preserve all other PII classes and non-argv spans
  -> Candidate/Detection tokenization pipeline
```

The same validator is applied to decoded BIO spans and direct backend detections so both NER entry paths behave consistently.

## Tests

Added regression coverage for:

- Idempotence: `redact -> restore -> redact` stays byte-stable for `icalBuddy`, `eventsToday`, `AppleScript`, the plan-my-day `osascript` AppleScript string, and already-stable argv tokens (`cal`, `ls`, `-l`, `~`).
- PII safety: `Dr. Schmidt`, `alice@example.invalid`, `/Users/alice/project`, `Workspace`, and `OpenAI` still pseudonymize and restore correctly.

## Privacy statement

This does NOT weaken PII redaction. The suppression applies only to model-origin `Name`/`Organization` spans that are shaped like command/argv identifiers. Emails, names, usernamed paths, organization tokens, deterministic detectors, and fail-closed behavior remain covered and unchanged.

## Gates

Green locally:

- `cargo fmt --all -- --check`
- `cargo test -p gaze-recognizers`
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `cargo test --workspace --all-features --no-fail-fast`
- `cargo run -p xtask -- readme-version-check`
- `cargo run -p xtask -- symmetric-potemkin`
- `cargo run -p xtask -- class-map-override-safety`
- `cargo run -p xtask -- recognizer-composition-validator`
- `cargo run -p xtask -- no-tenant-knowledge`
- `cargo run -p xtask -- bundle-tokenization-drift`
- `cargo run -p xtask -- fixture-citation-lint`
- `cargo run -p xtask -- ci-feature-matrix`

### DIFF
--- a/crates/gaze-mcp-core/tests/ui/tool_ctx_no_external_constructor.stderr
+++ b/crates/gaze-mcp-core/tests/ui/tool_ctx_no_external_constructor.stderr
@@ -9,8 +9,8 @@ error[E0624]: associated function `new` is private
    |     pub(crate) fn new(audit_session_id: &'a str) -> Self {
    |     ---------------------------------------------------- private associated function defined here
 
-error[E0599]: no associated function or constant named `new` found for struct `ToolCtx<'a>` in the current scope
+error[E0599]: no function or associated item named `new` found for struct `ToolCtx<'a>` in the current scope
   --> tests/ui/tool_ctx_no_external_constructor.rs:12:25
    |
 12 |     let _ctx = ToolCtx::new(_session, Value::Null, Ulid::new(), "tool", "principal");
-   |                         ^^^ associated function or constant not found in `ToolCtx<'_>`
+   |                         ^^^ function or associated item not found in `ToolCtx<'_>`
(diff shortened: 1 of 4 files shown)


Naoray (commented):
BLOCKER: GitHub reports PR #302 as blocked because the required DCO check is failing on the current rebased commits. Required fix: add valid Signed-off-by trailers to the PR commits, or explicitly authorize a follow-up history rewrite to add DCO signoffs, then rerun checks. Local privacy review and the full requested gaze gate suite passed, but I am not squash-merging while a required status check is red.

Issue #360, opened 2026-07-06
Collision-family class rename (custom:iban -> custom:family:payment-card-or-iban) silently breaks existing policies — IBANs leak


Naoray:
Found via downstream integration testing in CertaMesh/gaze-laravel (see CertaMesh/gaze-laravel#141) while refreshing policy behavior gates against the pinned binary v0.11.3.

## Symptom

The core rulepack now emits IBAN detections under the collision-family class `custom:family:payment-card-or-iban` instead of `custom:iban`:

```
$ printf 'IBAN DE89370400440532013000' | gaze clean --policy <default-tokenize probe>
class: "Custom:family:payment-card-or-iban", raw: "DE89370400440532013000"
```

Any policy written against the previously documented class names (`custom:iban`, `custom:credit_card`) with a `preserve` default now lets IBANs flow through **unredacted** — a silent never-leak violation for downstream consumers:

```
$ printf 'IBAN DE89370400440532013000' | gaze clean --policy resources/policy.toml
clean_text: "IBAN DE89370400440532013000"   # leak
```

## Questions / asks

1. Is the collision-family class name (`custom:family:payment-card-or-iban`) a **stable contract**, or an internal mechanism that leaked into the emitted class? Downstream policies need a name they can pin.
2. If the rename is intentional, it is a breaking change for every published policy — it should be called out in the changelog/upgrade notes, and ideally the old class names should still match (alias) for at least one deprecation cycle.
3. Consider a safer failure mode: when a rulepack emits a class no policy rule matches, a `preserve` default silently leaks. A warning on unmatched detection classes (or an opt-in strict mode) would surface this at policy-load time.

Downstream quickfix being applied in gaze-laravel: add `[[rule]] class = "custom:family:payment-card-or-iban" action = "tokenize"` to the published policy.

Naoray:
## Triage verdict

**The family class is an intentional, stable contract — but the leak you hit is a real defect in how the policy layer handles it, and the rename reached you undocumented because a CI forcing function was dormant. Fix + prevention in #362.**

Answers to your three questions:

[...]

PR #362, opened 2026-07-06
fix(policy): warn when collision-family fallback classes would silently leak (#360)

Naoray:
Fixes the silent-leak half of #360 (does not close it; triage verdict is commented on the issue).

## Triage verdict

The collision-family class `custom:family:payment-card-or-iban` is an **intentional, documented fail-closed mechanism** ([anchor-resolution.md](https://github.com/CertaMesh/gaze/blob/main/docs/explanation/detection/anchor-resolution.md)), not a leaked internal. But the report exposed a real never-leak defect *around* it, plus the process failure that let the change reach adopters undocumented.

## Root cause (five-whys)

1. **Why did IBANs leak downstream?** The policy tokenizes `custom:iban` with a `preserve` default; detections now carry `custom:family:payment-card-or-iban`, which matches no rule and falls through to `preserve` (`action_for`, `pipeline.rs`). Fail-closed at detection, fail-**open** at policy.
2. **Why did detections carry the family class?** `iban.structural` declares `mandatory_anchor = "iban"`. The anchor cue words (`IBAN`, `Account`, …) ship in the `locale-en`/`locale-de` rulepacks — not in `core`. With `bundled = ["core"]` the cue is never registered, so **every** IBAN takes the family fallback; `custom:iban` is de-facto unreachable in that config. (`[locale].active` only orders the fallback chain; it does not load a pack.)
3. **Why didn't adopters know?** The observable class change landed in v0.8 (`8ab9daf`, core/core-extended unification) with no CHANGELOG entry, no upgrade note, no deprecation path.
4. **Why didn't our gates force that?** `bundle-tokenization-drift` *did* capture the change — the committed snapshot contains the family class — but CI runs the gate without `--verify-ack`, so regenerating the baseline was enough to merge. The ack mechanism (drift-ack comment + CHANGELOG line) existed since v0.4.6 and was even used for the v0.7 spike (`snapshot_ack.rs` line 4), but nothing enforced it.
5. **Why does the policy layer fail open at all?** Unmatched classes default to `Preserve` with no diagnostics — reasonable for forward-compat, but invisible when the unmatched class is a *fallback* the detection layer emits deliberately.

## Changes (prevention system, three layers)

| Layer | Change |
|---|---|
| **Runtime guard** | `gaze clean` warns on stderr for every mandatory-anchor family class an active recognizer can emit that the policy leaves to a non-protective default — naming the exact `[[rule]]` to add. Emitted only on successful cleans so error envelopes stay single-line JSON. Library adopters get the same list from the new `gaze_assembly::uncovered_collision_family_classes`. |
| **Contract docs** | `docs/reference/policy.md` gains "Collision-family fallback classes (avoid a silent leak)": family classes are a pinnable contract, `[locale].active` vs rulepack loading, covering-rule pattern. |
| **CI forcing function** | `bundle-tokenization-drift` now runs with `--verify-ack` (checkout gets the `fetch-depth: 0` it needs). Any future snapshot change — including emitted class renames — cannot merge without a `// drift-ack:` comment and a CHANGELOG line. |

Scoped to mandatory-anchor families (systematic fallback emitters); rare precedence-tie families are excluded to keep the warning low-noise. Aliasing the family class back to `custom:iban` was considered and rejected: the family token is emitted precisely when the span is *ambiguous*, so re-narrowing it would fake certainty and weaken axis 4 (trust/auditability).

## North-star axes

- **Axis 1 (never leak):** the default `core`-only + `preserve`-default configuration now warns instead of leaking silently.
- **Axis 4 (trust):** class-contract changes become auditable events (drift-ack + CHANGELOG) instead of silent snapshot refreshes.
- **Axis 5 (ergonomics):** the warning names the exact rule to paste.

## Tests

- `gaze-assembly`: 3 unit tests on `uncovered_collision_family_classes` (flags the leak config, silent on protective default, silent when family covered).
- `gaze-cli`: 2 end-to-end tests reproducing the issue's exact config — warning + leak surfaced, and covered-family silence.
- Full suites green: `gaze-assembly` 44, `gaze-cli` `cli_pipe` 104. `cargo fmt --check`, `clippy` (no new warnings), `bundle-tokenization-drift --verify-ack`, `fixture-citation-lint`, `readme-version-check` all pass under the pinned 1.96.0 toolchain.

https://claude.ai/code/session_016dVthtTfAGD8e2opbJ85Sj

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -136,6 +136,10 @@ jobs:
           docker-images: true
           swap-storage: false
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          # bundle-tokenization-drift --verify-ack diffs snapshots against
+          # origin/main...HEAD; a shallow clone silently degrades it to a no-op.
+          fetch-depth: 0
       - uses: dtolnay/rust-toolchain@29eef336d9b2848a0b548edc03f92a220660cdb8 # 1.96.0
         with:
           toolchain: 1.96.0
@@ -176,7 +180,11 @@ jobs:
       - name: xtask no-tenant-knowledge
         run: cargo run -p xtask -- no-tenant-knowledge
       - name: xtask bundle-tokenization-drift
-        run: cargo run -p xtask -- bundle-tokenization-drift
+        # --verify-ack: any committed snapshot change (e.g. an emitted class
+        # rename) must carry a `// drift-ack:` comment and a CHANGELOG line —
+        # the v0.8 `custom:iban` -> `custom:family:payment-card-or-iban` change
+        # merged silently without this and leaked downstream (issue #360).
+        run: cargo run -p xtask -- bundle-tokenization-drift --verify-ack
       - name: xtask family-policy-table-coherence
         run: cargo run -p xtask -- family-policy-table-coherence
       - name: xtask locale-cue-bundle-coherence
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`gaze clean` now warns when a collision-family fallback class would silently
+  leak** (#360). When an active recognizer can emit a `custom:family:<family>`
+  class (for example `custom:family:payment-card-or-iban` for an anchor-less
+  IBAN) and the policy leaves that class to a non-protective default, the CLI
+  prints a `warning:` to stderr naming the exact rule to add. Rust adopters get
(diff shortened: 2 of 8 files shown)


Naoray:
Subagent review pass found one critical defect, fixed in d152485: the coverage check ignored rule order — the runtime matcher is first-match-wins with an unconditionally-matching `default`, so a covering family rule pasted *after* the default rule is dead code, yet the checker counted it as covered and silenced the warning while the IBAN kept leaking. Coverage is now restricted to rules declared before the first `default` rule; the warning and docs now say to add the covering rule **before** the default rule; regression tests added at both the `gaze-assembly` unit and `gaze-cli` end-to-end layers (shadowed rule → leak persists → warning keeps firing). Second reviewer verified all regex/docs claims on the sibling PR #363 with no findings.

PR #395, opened 2026-07-16
[agent] docs: index benchmarks and show metric goals

Naoray:


Naoray:
Superseded by a fresh exact-tree commit with the required DCO trailer; no amend or force-push was used.

PR #404, opened 2026-07-25
[agent] Add a corpus-informed scheme/www-anchored URL recognizer to the core rule floor

Naoray:
## Summary

Adds a corpus-informed scheme/`www.`-anchored URL recognizer to the core rule floor. Before this change the deterministic rule floor was **100% blind to URLs** — 0 fully covered *and* 0 overlapped of 276 gold entities — and URL was the largest measured leak bucket in the no-OPF benchmark (todo #2254).

The recognizer is `safe_default` and locale-`global`, so it fires in all three benchmark cells and for default adopters, not only under the benchmark's configuration.

The rule shape was chosen by measurement over the corpus, not by intuition:

- **83.0% of gold URL spans (229/276) and 87.7% of the leaked URL bytes (6,322/7,209) carry an explicit `https://`, `http://`, or `www.` anchor.** Only the bare-host tail (47 spans, 887 bytes, 12.3%) lacks one, and that tail is deliberately left out.
- The committed A4 negative corpus (1,024 documents, 512 en / 512 de) contains **zero** scheme-anchored and **zero** `www.`-anchored URL-shaped matches, so the anchored form is provably safe against the negative corpus that #2254 item 3 requires it to clear. Bare-host shapes appear in 97 A4 documents, which is precisely why the bare-host tail is excluded.
- The boundary rule (give back trailing punctuation vs. run to whitespace) was also chosen by measurement: identical coverage, false positives in 12 documents versus 107.

## Measured result (full profile, complete EN/DE holdout + 1,024 A4 negatives)

Baseline: `origin/agent/current-gaze-benchmark` @ `3163b56`, which reproduces the committed schema-v3 scorecard exactly (URL 7,209 / 6,223 / 4,847 leaked bytes per cell).

| | rule-floor-extended | pass2-ner | full-stack-kiji-resolve ⚠ |
|---|---:|---:|---:|
| **URL leaked bytes** | 7,209 → **824** (−6,385, −88.6%) | 6,223 → **2,357** (−3,866, −62.1%) | 4,847 → **1,922** (−2,925, −60.3%) |
| **URL fully covered** | 0 → **232** | 0 → **138** | 0 → **105** |
| URL overlapped | 0 → 232 | 104 → 242 | 100 → 194 |
| URL wholly missed | 276 → **44** | 172 → **34** | 120 → **25** |
| **Total leaked bytes** | 112,453 → **106,068** (−6,385) | 44,561 → **40,695** (−3,866) | 30,693 → **27,728** (−2,965) |
| **Zero-leak documents** | 1,024 → 1,024 (+0) | 1,054 → **1,067** (+13) | 908 → **919** (+11) |

⚠ See the population caveat under deviation 3 below.

## Reproduction evidence

The full profile was run **twice** on the candidate with `--compare-baseline`. **1,117 integer leaves compared across all three cells; zero differences.** Every integer in the tables above reproduces exactly.

Gates under the repo-pinned toolchain (rustc 1.96.0): `cargo fmt --all -- --check`, `cargo clippy --workspace --all-features --all-targets -- -D warnings`, and `cargo test --workspace --all-features` all pass with **zero** test failures.

## Contract intact

No change to any reversibility, manifest, or trace counter:

| | rule-floor-extended | pass2-ner | full-stack-kiji-resolve |
|---|---|---|---|
| Restore exact documents | 2910 → 2910 | 2910 → 2910 | 2250 → 2250 |
| Manifest valid documents | 2910 → 2910 | 2910 → 2910 | 2250 → 2250 |
| Manifest integrity errors | 0 → 0 | 0 → 0 | 0 → 0 |
| Redact actions | 0 → 0 | 0 → 0 | 0 → 0 |
| Strict would-reject documents | 0 → 0 | 0 → 0 | 1,268 → **1,264** (improved) |
| Completed / failed-closed | 2910 / 0 (+0) | 2910 / 0 (+0) | 2250 / 660 (+0) |

Manifest spans and final protection trace items rise in step with tokenize actions (+241 / +140 / +257), which is the expected consequence of emitting more tokens; spans, trace items, and tokenize actions stay equal to one another in every cell.

## #2254 item 8 deviations

**This PR does not clear #2254 item 8.** Strict reduction in leaked bytes: yes, decisively. No regression in any other canonical counter: **no**. All three deviations are enumerated below rather than summarized, and the merge decision belongs to the user.

For context, the release-readiness verdict fails on **both** sides — the base is not release-ready either — and the candidate is strictly better than the base on every leak gate.

### 1. False-positive increase — accepted, disclosed

| | FP bytes | FP documents |
|---|---:|---:|
| rule-floor-extended | 5,038 → 5,353 (**+315**) | 209 → 221 (**+12**) |
| pass2-ner | 27,681 → 27,775 (**+94**) | 1,943 → 1,944 (**+1**) |
| full-stack-kiji-resolve | 92,106 → 92,847 (**+741**) | 1,857 → 1,857 (+0) |

These are the **full-scale** numbers and are the ones that must be quoted; earlier quick-profile figures (+44 B / +1 doc deterministic, +151 B / 0 docs kiji) sampled only 256 of 2,910 documents and understate the cost.

The increase is caused by 9 anchored URLs that the corpus never labelled. Accepted on the following basis:

- The trade is **20:1** (rule-floor), **41:1** (pass2), and **4:1** (kiji); overall **11.5:1** — 13,216 leaked bytes removed for 1,150 false-positive bytes added.
- Scaling was checked rather than assumed: from quick (256 docs) to full (2,910 docs, ~11×), FP grew ~7× while leak reduction grew ~10×. The FP cost grew **slower** than the win, and the rule-floor ratio improved from 14:1 to 20:1.
- AGENTS.md axis 1 (never leak) beats axis 5 (adopter ergonomics), and correctness axes 1–4 beat performance.
- No principled narrowing exists. Fitting the rule to those 9 unlabelled instances was explicitly refused: it would be corpus overfitting, the same failure mode that caused a competing Kiji stack to be dropped from this wave. A clean FP column bought by special-casing unlabelled instances is worth less than an honest one.

### 2. Kiji residual/post-policy suspects 170 → 172 (+2) — confounded, not attributable

Bounded attribution, aggregate/non-PII telemetry only. The question asked was whether the 2 additional residual suspects sit in documents inside the membership-swap delta or in documents common to both scored sets. **It cannot be resolved from available telemetry, and the evidence is exactly consistent with confounding by the population swap in deviation 3:**

- At full scope, exactly **2** documents flip from zero-suspect to non-zero (`post_policy_zero_suspect_documents` 2,122 → 2,120) and suspects rise by exactly **2**; so each contributes exactly one suspect.
- At full scope, exactly **2** documents change language membership in the kiji cell (en 1,488 → 1,486, de 762 → 764) while the total stays 2,250. The arithmetic matches the swap one-for-one.
- Against a systematic mechanism: at the quick profile the recognizer is demonstrably active in the kiji cell (+41 tokenize actions), yet residual suspects do **not** move at all (20 → 20). If new URL tokens systematically changed what the observer-only post-clean pass sees, the quick profile should show it.
- There is no per-document or per-language residual telemetry, so the two documents cannot be located directly.

Disclosed as confounded pending #2414. No fix attempted.

### 3. `gold_entities_match` population failure — kiji cell is not provably like-for-like

The regression verdict fails `gold_entities_match` in the kiji cell: candidate 10,665 versus baseline 10,668, together with `pii_utf8_bytes_match` 93,985 versus 94,037.

`gaze_bench_score.py:861` computes `self.entities += len(document.spans)` from corpus ground truth and never reads `predictions`, so **for a fixed document set gold is detector-invariant**. This is not a gold-derivation defect, not entity segmentation, and not overlap attribution. The only available mechanism is a **scored-set membership swap at constant cardinality**: one document begins failing closed while a different one stops, preserving counts but not identity.

Confirmed at both scales, in both directions, with `documents` and `failed_closed_documents` identical on both sides:

| profile | documents | failed-closed | gold entities |
|---|---:|---:|---:|
| quick | 199 → 199 | 57 → 57 | 934 → **940** |
| full | 2,250 → 2,250 | 660 → 660 | 10,668 → **10,665** |

Consequences, stated precisely:

- **`rule-floor-extended` and `pass2-ner` are strictly like-for-like and quotable without caveat.** Both have `failed_closed_documents = 0` and score the complete corpus: documents 2,910 and gold 14,719, identical on both sides. There is no population to swap.
- **The `full-stack-kiji-resolve` cell is not provably like-for-like** and every figure from it in this PR carries the #2414 caveat.

For context — and as context only, not as a dismissal of the caveat — the swap is 3 gold entities out of 10,668 and 1 URL entity out of 220, against a 2,925-byte URL reduction, so it cannot plausibly account for the measured delta. The caveat stands regardless: a delta over a population that is not provably identical is not a like-for-like delta.

Tracked as todo #2414 (scored population reported as a cardinality, so membership swaps are invisible).

### 4. Class coverage losses in the kiji cell — found after this PR was drafted

This section was added after the fact and it corrects an omission in the original body. The
analysis behind the first draft read only the `URL` label. Applying a like-for-like discriminator
across **every** label — gold entity count unchanged **and** `fully_covered` or `overlapped`
decreased — shows this change also removed protection from three classes it was not aiming at:

| class | entities | fully_covered | overlapped | leaked bytes |
|---|---:|---|---|---:|
| SURNAME | 1,202 (**constant**) | 1,196 → **1,194** (−2) | 1,199 → 1,198 (−1) | +5 |
| LICENSEPLATENUM | 158 (**constant**) | 0 → 0 | 48 → **47** (−1) | +1 |
| TAXNUM | 149 (**constant**) | 0 → 0 | 10 → **9** (−1) | +8 |

All three sit in `full-stack-kiji-resolve`; `rule-floor-extended` and `pass2-ner` are clean.

**Why the entity-constant qualifier matters, and how to verify it independently.** The scorer
merges every prediction class-agnostically before computing coverage (`gaze_bench_score.py`,
`MetricAccumulator.add`: `predicted = merge_intervals(... for span in predictions)`, no class
filter). So on an unchanged population, an entity leaving the `overlapped` bucket means **no
prediction of any class intersects it any more** — the bytes are raw on the wire, not merely
attributed to the wrong class. A wrong-class token would still count as both overlapped and
covered. Anyone can re-derive the table above from `diagnostics.json`'s `per_label` block in the
two committed runs.

**Scale, stated honestly.** 2 fully-covered entities lost, +14 leaked bytes across the three
classes, against this change's −6,385 URL bytes at the rule floor. The ratio is overwhelmingly
favourable. It is disclosed because a net-favourable trade is still a trade, and because a
credential-class instance of this same mechanism was later found on a different branch and judged
serious enough to hold that work.

**Mechanism.** Every known instance is confined to `full-stack-kiji-resolve`, whose safety net
runs in resolve mode against already-tokenized output — so adding tokens upstream changes the
input that pass sees, and therefore what it predicts. It is not a resolver or collision-family
decision: a collision-precedence fix aimed at this on another branch changed **zero** of 1,117
integers. Tracked as todo #2420.

**Detection.** A `class_coverage_loss` regression gate implementing exactly this discriminator now
exists; replayed against this change it fires on all three classes above. It did not exist when
this PR was drafted, which is why the losses went unnoticed.

## Not caused by this work

Both deterministic cells report 5 `restore_decision_failures` (`restore_success_decisions` 2,905 of 2,910). These counts are **byte-identical in the base run** and predate this change entirely. Filed separately as todo #2415 so this PR's deviation list is not contaminated with a defect that is not its own.

## Scope

No change to the stable public API, the schema-v3 wire/cell/verdict shape, the root `Cargo.toml`, or the root `Cargo.lock`. No OPF, OCR, document, structured, or streaming scope. Committed baseline evidence is untouched; the comparison baseline used here is a local, gitignored file under `target/bench-data/`.

Resolves solo todo #2254.

### Status

This body presents evidence; it does not recommend a merge decision. Under `#2254` item 8 the
change is a strict, large reduction in leaked bytes with **four** categories of deviation — the
accepted false-positive increase, the confounded kiji residual +2, the population/`gold_entities_match` failure, and the class
coverage losses documented above. The merge decision rests with the maintainer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Scheme- and `www.`-anchored URL detection at the deterministic rule floor**
+  (todo #2254). The new `url.anchored` recognizer in the embedded `core` bundle
+  tokenizes `http://`, `https://`, and `www.`-prefixed URLs as
+  `custom:url`, covering the whole span rather than a fragment. It is
+  `safety_tier = "safe_default"` with `locales = ["global"]`, so it is active for
+  every adopter of the default bundle, not only for configurations that
+  auto-activate locale-gated recognizers.
+
+  Measured against the EN/DE benchmark holdout: those two anchors carry 232 of
+  276 gold URL spans and 6,385 of 7,209 leaked URL bytes (88.6%), while matching
+  nothing at all across the 1,024 documents of the committed A4 negative corpus.
+  Before this recognizer the deterministic rule floor was completely blind to
+  URLs — 0 of 276 spans covered and 0 overlapped.
+
+  **Bare-host URLs are deliberately out of scope.** Shapes with no scheme and no
+  `www.` prefix (`example.invalid/orders`) are the remaining 47 spans / 887 bytes
+  / 12.3% of the bucket, and 97 of the 1,024 committed negative documents contain
+  bare-host shapes, so no bare-host rule can clear the negative gate.
+
+  **Documentation, repository, and example URLs are tokenized.** This is
+  intentional: 16 of the 232 gold spans the rule covers are themselves
+  reference-host shaped, so the corpus treats a reference URL inside a
+  data-owner document as PII to protect. Tokens stay restorable through the
+  manifest, so an over-tokenized public URL is a recoverable ergonomics cost
+  (axis 5) while an under-tokenized private one is a leak (axis 1).
+
+### Changed
+
+- `[bundle-tokenization-drift]` snapshot for bundle `core` regenerated: the new
+  `url.anchored` recognizer adds one `custom:url` detection to the drift corpus
+  (11 -> 12 detections). No existing detection changed class, span, or shape.
+
 ## [0.12.0] - 2026-07-06
 
 ### Added
--- a/crates/gaze-recognizers/embedded/core.toml
+++ b/crates/gaze-recognizers/embedded/core.toml
@@ -717,3 +717,44 @@ base = 0.70
 priority = 70
 
 [recognizers.token]
+
+# URL detection is anchored on an explicit scheme (`http://`, `https://`) or a `www.` host
+# prefix. Measured against the EN/DE holdout, those two anchors carry 232 of 276 gold URL spans
+# and 6,385 of 7,209 leaked URL bytes (88.6%), while matching NOTHING in all 1,024 documents of
+# the committed A4 negative corpus. Bare-host URLs (`example.invalid/path`, no scheme, no `www.`)
+# are the remaining 12.3% and are deliberately NOT matched: 97 of the A4 negative documents
+# contain bare-host shapes, so no bare-host rule can clear the negative gate. See todo #2254.
+#
+# Trailing-boundary rule: the match runs to the next whitespace and then gives back a run of
+# sentence and markup punctuation, so `.`, `,`, `;`, `:`, `!`, `?`, `)`, `]`, `}`, `>`, `"`, `'`
+# never end a match. This is measured, not stylistic: running to whitespace without giving those
+# back covers exactly the same 232 spans and 6,385 bytes but produces non-gold matched bytes in
+# 107 holdout documents instead of 12 — sentence-final punctuation is the dominant
+# false-positive driver for an anchored URL rule.
+[[recognizers]]
+id = "url.anchored"
+safety_tier = "safe_default"
+class = "custom:url"
+enabled = true
+locales = ["global"]
+
+[recognizers.match]
+kind = "regex"
+# Adding this recognizer changes the committed core-no-policy bundle tokenization snapshot
+# (11 -> 12 detections). The `// drift-ack:` acknowledgement the gate requires lives in
+# crates/gaze-recognizers/tests/url_anchored.rs; the rationale lives in CHANGELOG.md.
+pattern = '''(?i)(?-u:\b)(?:https?://|www\.)[^\s]*[^\s.,;:!?)\]}>"']'''
+
+[recognizers.scoring]
+# Above the 0.70 heuristic floor: an explicit scheme or `www.` prefix is structural evidence,
+# not a guess. Priority stays below `email.global` (90) so an explicit address inside a URL
+# keeps its Email class rather than being absorbed; note that `PiiClass::Custom` sits at
+# class-priority 50 in the resolver, so Email wins containment regardless.
+base = 0.75
+priority = 85
+
+[recognizers.token]
+
+[recognizers.source]
+origin = "project-authored"
+license = "Apache-2.0"
--- a/crates/gaze-recognizers/src/lib.rs
+++ b/crates/gaze-recognizers/src/lib.rs
@@ -66,7 +66,7 @@ mod tests {
         let core = embedded("core").expect("core rulepack");
         let rulepack = Rulepack::load(RulepackSource::Embedded(core)).expect("valid core");
 
-        assert_eq!(rulepack.recognizers.len(), 29);
+        assert_eq!(rulepack.recognizers.len(), 30);
         assert_eq!(rulepack.recognizers[0].id, "email.global");
         assert_eq!(rulepack.recognizers[1].id, "email.header.name");
         assert_eq!(rulepack.recognizers[2].id, "email.header.name.paren");
@@ -101,7 +101,7 @@ mod tests {
         let rulepack =
             Rulepack::load(RulepackSource::Embedded(core_extended)).expect("valid core-extended");
 
-        assert_eq!(rulepack.recognizers.len(), 29);
+        assert_eq!(rulepack.recognizers.len(), 30);
         assert!(rulepack
             .recognizers
             .iter()
--- a/crates/gaze-recognizers/tests/url_anchored.rs
+++ b/crates/gaze-recognizers/tests/url_anchored.rs
@@ -1,0 +1,465 @@
+//! Regression fixtures for the `url.anchored` core recognizer (solo todo #2254).
+//!
+//! EVERY positive fixture here encodes a structural shape that was MEASURED in the Dataiku EN/DE
+//! holdout, not a phrasing invented alongside the implementation. The previous attempt at this
+//! bucket (todo #2412) shipped 962 green tests for cue phrasings — "my profile is <url>" — that
+//! occur in 5 of 276 gold spans, and produced a zero real-corpus delta. The measured distribution
+//! of the 232 gold URL spans this rule covers is:
+//!
+//! | dimension      | counts                                            |
+//! |----------------|---------------------------------------------------|
+//! | anchor         | scheme 169, `www.` 63                             |
+//! | shape          | host-only 149, has-path 83                        |
+//! | host labels    | 2 labels 57, 3 labels 154, 4 labels 21            |
+//! | query string   | 1                                                 |
+//! | reference host | 16 (docs./github/wiki/example. shapes — all GOLD) |
+//!
+//! Every row above has at least one fixture below, and the trailing-boundary cases exist because
+//! the boundary rule was chosen on measured false-positive counts (107 documents vs 12).
+//!
+//! Fixture values are synthetic and use reserved `.invalid` hosts only.
+
+use gaze::Context;
+use gaze::{
+    Action, CleanDocument, DictionaryBundle, LocaleChain, LocaleTag, PiiClass, Pipeline,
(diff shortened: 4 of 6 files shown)


Naoray:
## Orchestrator verification — independent re-derivation, then merge

Every headline integer in this body was re-derived from the two committed scorecard/diagnostics
JSONs without reading the PR's own analysis, and the full-label like-for-like discriminator was
re-implemented from scratch rather than replayed through the `#405` gate.

**Provenance of the artifacts checked first.** `baseline.json` / `agent-bench-baseline-pristine`
carries `gaze.revision = 3163b56`, `dirty = false`, profile `full`, 2,910 documents; the candidate
carries `662cde4`, `dirty = false`, same profile and document count. The base branch has since
advanced `3163b56 → 0e9d03a`, and that advance is `scripts/bench/gaze_bench_score.py` +
`scripts/bench/test_class_coverage_gate.py` only (+254/-0, no crate touched), so the measurement
basis is not stale.

**Cell totals — reproduce exactly.**

| | rule-floor | pass2-ner | kiji-resolve |
|---|---|---|---|
| leaked bytes | 112,453 → 106,068 | 44,561 → 40,695 | 30,693 → 27,728 |
| FP bytes | 5,038 → 5,353 (+315) | 27,681 → 27,775 (+94) | 92,106 → 92,847 (+741) |
| FP documents | 209 → 221 (+12) | 1,943 → 1,944 (+1) | 1,857 → 1,857 (+0) |
| gold entities | 14,719 → 14,719 | 14,719 → 14,719 | 10,668 → **10,665** |

**Reversibility — intact in all three cells.** `restore_exact_documents` constant (2,910 / 2,910 /
2,250) at `restore_exact_rate = 1.0`; all six `manifest_integrity_errors` sub-counters 0 → 0;
`manifest_valid_documents` constant; `redact_actions` 0 → 0 (no one-way primitive enters the path);
`failed_closed_documents` 660 → 660. `manifest_spans`, `tokenize_actions` and
`final_protection_trace_items` stay equal to one another in every cell while rising in step
(+241 / +140 / +257). `strict_would_reject_documents` improves 1,268 → 1,264.

**Deviation 4 confirmed, and confirmed complete as an entity-bucket claim.** The independent sweep
over all 29 labels × 3 cells returns exactly the three disclosed losses and no others:

| cell | label | gold | fully_covered | overlapped | leak Δ |
|---|---|---:|---|---|---:|
| kiji | SURNAME | 1,202 const | 1,196 → 1,194 | 1,199 → 1,198 | +5 |
| kiji | LICENSEPLATENUM | 158 const | 0 → 0 | 48 → 47 | +1 |
| kiji | TAXNUM | 149 const | 0 → 0 | 10 → 9 | +8 |

17 kiji-cell labels are not evaluable because gold moved — including `URL` itself (220 → 219), which
is why every kiji figure here correctly carries the `#2414` caveat. `rule-floor-extended` and
`pass2-ner` are clean and fully evaluable on all 29 labels.

### One finding this body does not contain

`PASSWORD` in the kiji cell: gold constant at 153, `fully_covered` and `overlapped` **both
unchanged**, and `leaked_utf8_bytes` **+1**. It is one byte on a credential class, but it is a real
byte-level movement that the disclosed table does not list — because the `#405` discriminator is
entity-bucket-based, so a partially-covered entity that loses coverage without leaving its bucket is
invisible to it. Filed as a follow-up against the gate, not against this change: the gate needs a
byte-level arm (`gold constant AND leaked_utf8_bytes increased`) alongside the bucket arm.

### Verdict: merge

- **Axis 1 decides it.** The deterministic rule floor was 100% blind to URLs — 0 of 276 covered
  *and* 0 of 276 overlapped — and is now 232 covered with 88.6% of the URL leak bucket closed. The
  aggregate trade is 13,216 leaked bytes removed for 1,150 false-positive bytes added, and the three
  disclosed coverage losses plus the `PASSWORD` byte total **+15 bytes** against **−13,216**.
- **Axis 2 is untouched**, verified counter by counter above.
- **Axis 4 holds**: reproduced twice at 1,117 integer leaves with zero differences, and reproduced a
  third time here from the artifacts by a different implementation of the discriminator.
- The base branch also fails release readiness. The candidate is strictly better on every leak gate,
  and the residual deviations are attributed to `#2414` (population reported as cardinality) and
  `#2420` (kiji resolve-mode input sensitivity) — both systemic, both pre-existing, neither caused
  by this recognizer.
- The false-positive column was **not** cleaned by special-casing the 9 unlabelled anchored URLs.
  That refusal is the right call and is what makes the rest of the numbers trustworthy.

The regex was reviewed independently as well: the `(?-u:\b)` ASCII-word-boundary prefix prevents
matching inside identifiers (`xwww.host` does not match), the give-back tail class cannot end a match
on sentence or markup punctuation, `www.` and `https://` alone do not match, and the
class-priority claim in the recognizer comment is accurate — `resolver.rs:387` puts `Email` at 90
against `PiiClass::Custom(_)` at 50, and `should_replace_containment` prefers the higher class
priority, so an explicit address inside a URL keeps its `Email` class.

PR #435, opened 2026-08-17
[agent] fix(resolver): one arbitrate() with a shared ladder; decided_by always names the deciding tier

Naoray:
## Summary

`crates/gaze/src/resolver.rs` carried three identical win/lose/merge bookkeeping
blocks, two byte-identical wrappers of one comparator, and the base conflict
ladder written out twice. It decided each overlap with a forward comparator call
and then reused a *second, reverse* call as the winner's `decided_by` label —
sound only if the comparator is antisymmetric. The mandatory-anchor rung is not:
it inspects its first argument alone and returns `AnchoredContext` regardless of
the rival. Two audit-integrity defects followed.

Three commits, in order:

1. **`614aae8` — pure refactor, output-identical.** One `Overlap{Exact,
   Containment, Partial}` classification, one `arbitrate() -> Arbitration{Merge,
   Family, CandidateWins(tier), ExistingWins(tier)}`, one bookkeeping copy, one
   shared `compare_base_ladder`. No behavior change of any kind.
2. **`d6def03` — `decided_by` label fix** (audit-row change, separately
   reviewable).
3. **`6a18c99` — mutation locks for the two invariants review found unlocked**
   (tests only; no production change). See [Mutation locks](#mutation-locks).

Detection outcomes are unchanged throughout: which spans get tokenized, their
classes, and the restore contract are byte-identical. Only `decided_by` (and the
`ambiguity_record` derived from it downstream) moves, and only in commit 2.

## The defects (commit 2)

**1. `AnchoredContext` stamped on a ladder-decided overlap.** When an anchored
incumbent kept its span on `Score`, `SpanLength`, or `RecognizerId`, the reverse
probe hit the anchor short-circuit and labelled the row `AnchoredContext` — an
audit row claiming a *missing-anchor family fallback* decided a conflict the base
ladder decided. Reachable in the shipped default: `iban.structural` declares
`mandatory_anchor = "iban"` in the embedded `core` bundle
(`crates/gaze-recognizers/embedded/core.toml`).

**2. Stale tier survived a full tie.** When every rung tied, the reverse probe
returned `None` and the incumbent kept whatever `decided_by` an *earlier* overlap
had left on it, instead of naming its own terminal rung.

Both violate axis 4 (trust — every token emission traceable to the rule that
actually decided it). Neither is a leak: no span or class outcome changes.

### Red → green

Regression tests were written first and run against commit 1 (the refactor
alone), where they fail with the exact mislabel:

```
test resolver::tests::anchored_incumbent_that_wins_on_score_is_labelled_score ... FAILED
test resolver::tests::anchored_incumbent_that_wins_partial_overlap_on_score_is_labelled_score ... FAILED
test resolver::tests::fully_tied_partial_overlap_never_keeps_stale_tier ... FAILED
test anchored_iban_beating_a_rival_on_score_logs_score_not_anchored_context ... FAILED

panicked at crates/gaze/src/resolver.rs:611:  left: AnchoredContext   right: Score
panicked at crates/gaze/src/resolver.rs:645:  left: AnchoredContext   right: Score
panicked at crates/gaze/src/resolver.rs:661:  left: None              right: RecognizerId
panicked at crates/gaze/tests/contracts.rs:918: left: AnchoredContext  right: Score
```

After `d6def03`: all four green. The `contracts.rs` case is end-to-end — a real
`iban.structural` recognizer with a registered `DeDe` anchor cue bundle, an
anchor present in the input, and a lower-score `digits.generic` rival — and
asserts the persisted audit row for both the winner and the losers.

## Proof the refactor is output-identical

The pre-refactor `insert_candidate` and its three comparators were kept verbatim
in a throwaway `legacy_diff` module and fuzzed against the new code
(**20,000 cases**, spans 0..24 × lengths 1..8 × 7 recognizers × 4 scores ×
3 priorities × 0–8 candidates, crossed with: family policy on/off, anchor context
on/off, and an `IBAN` cue placed at a random offset so `AnchorOutcome::Found`,
`Missing`, and `NotRequired` are all exercised):

- **Commit 1:** full `Vec<Candidate>` equality — identical on every case.
- **Commit 2:** every field except `decided_by` identical on every case (span,
  class, `recognizer_id`, `source`, `merged_sources`, `token_family`,
  `canonical_form`, `score` bitwise, `priority`). Strict equality fails only on
  `decided_by`, confirming the fuzz reaches the changed label rather than
  passing vacuously.

The module is throwaway scaffolding and is **not** in either commit; the
permanent replacement is the new proptest below.

Mutation-checked, not assumed: swapping the anchor tier to `Score` in the new
code makes the differential fuzz FAIL; dropping `remove_overlaps` and pushing
the loser instead of dropping it makes the new proptest FAIL. Both reverted.

## Mutation locks

Review ([#435 review comment](https://github.com/CertaMesh/gaze/pull/435#issuecomment-5320525115))
proved two entries in the invariants checklist below were **documented but not
locked**: the invariant survived its own mutation. Both mutations were
reproduced here before fixing, and both locks were then verified to fail under
the mutation and pass on restore. Commit `6a18c99`, tests only.

### 1. Family tie before the exact same-class merge (solo todo #2964)

**Mutation:** move the `family_tie_candidate` probe below the
`overlap == Overlap::Exact && existing.class == candidate.class` merge in
`arbitrate`.

**Before:** the entire `gaze-pii` suite stayed green (`--all-features`, all
targets), all three precedence-tie fixtures included. Cause: every existing
fixture pairs candidates of **different** classes, so the same-class merge those
fixtures are meant to preempt never runs.

**Lock:** `exact_span_same_class_precedence_tie_emits_family_candidate_not_merge`
— two equal-precedence variants of one collision family on the **same span** with
the **same class**, the only shape where the merge can swallow the tie.

```
test resolver::tests::exact_span_same_class_precedence_tie_emits_family_candidate_not_merge ... FAILED

assertion `left == right` failed: an exact same-class overlap must not merge ahead of the family tie
  left: Custom("tenant_doc")
 right: Custom("family:tenant-document")
```

### 2. Exact-span paths never call `remove_overlaps` (solo todo #2963)

**Mutation:** remove **both** `overlap != Overlap::Exact` guards.

**Before:** `prop_resolver_invariants` stayed green at `PROPTEST_CASES=2048`, and
so did the full suite.

**Root cause — the guards are output-invisible, not merely untested.** An
exact-overlap winner installs a span identical to the one it evicted, and the
resolved set is pairwise disjoint (property 1 of `prop_resolver_invariants`), so
`remove_overlaps` finds nothing to remove and returns an identical set. A
20,000-case differential over a corpus biased toward exact-span collisions,
digesting **every** `Candidate` field including `decided_by` and
`merged_sources`, produced `f340c66d90503b37` **both with and without both
guards**. No black-box test can ever fail on this mutation, so the locks use
`cfg(test)` instrumentation — the route the review explicitly authorised. (That
differential was throwaway scaffolding and is not committed; the in-repo basis
for the claim is the disjointness property.)

**Lock:** a `#[cfg(test)]` thread-local count of `remove_overlaps` entries, one
test per guarded arm, plus a positive control so neither zero-assertion can pass
vacuously:

| Test | Locks |
|---|---|
| `exact_span_family_tie_skips_overlap_removal` | the `Arbitration::Family` arm |
| `exact_span_candidate_win_skips_overlap_removal` | the `Arbitration::CandidateWins` arm |
| `non_exact_winner_enters_overlap_removal` | positive control — a widening winner **must** enter removal, so a counter stuck at zero cannot make the two above vacuous |

```
test resolver::tests::exact_span_family_tie_skips_overlap_removal ... FAILED
test resolver::tests::exact_span_candidate_win_skips_overlap_removal ... FAILED

assertion `left == right` failed: an exact-span family tie must not enter overlap removal
  left: 1
 right: 0
assertion `left == right` failed: an exact-span collision-policy win must not enter overlap removal
  left: 1
 right: 0
```

Removing **either** guard alone fails exactly its own test and no other, so the
two arms are locked independently rather than only as a pair.

**No production change.** `REMOVE_OVERLAPS_CALLS` and its increment are both
`#[cfg(test)]`, so the shipped resolver is byte-unchanged. The guards were
already correct; commit `6a18c99` adds coverage only.

### Test roster

| Suite | Result |
|---|---|
| `resolver::tests` unit tests | 9 pre-existing green unchanged, +3 label fixes, +4 mutation locks (17 total incl. `anchor_resolver`) |
| `contracts.rs` (`decided_by` assertions incl. both mandatory-anchor contracts) | 42 green unchanged, +1 new = 43 |
| `prop_manifest_invariants` | 4 green unchanged |
| `prop_resolver_invariants` (**new**) | 2 properties; 128 cases each by default (`PROPTEST_CASES` overrides), verified at 512 |
| `cargo test --workspace --all-features --no-fail-fast` | green (exit 0); no `kiji subprocess timed out` and no OPF `Timeout` observed |
| `cargo test -p gaze-pii --all-features` (at `6a18c99`) | green — 192 + 43 + 4 + 2 + … all suites |
| `cargo clippy --workspace --all-features --all-targets -- -D warnings` | clean |
| `cargo fmt --all -- --check` | clean |
| `cargo run -p xtask -- family-policy-table-coherence` | pass |
| `cargo run -p xtask -- symmetric-potemkin` | pass |

`crates/gaze/tests/prop_resolver_invariants.rs` is new and asserts at the public
resolver boundary: the resolved set is pairwise disjoint and sorted, every
resolved boundary comes from an input boundary, the set never grows, and the
result is invariant under input permutation — with and without a family policy
(including an equal-precedence pair that emits a family-level token). **It was
written first and confirmed green against `main`'s resolver** before the refactor
existed, so it is a true invariant of the boundary and not a description of the
new implementation.

## Invariants checklist

- [x] **Family-tie-before-merge order** — `arbitrate` checks
      `family_tie_candidate` before the same-span same-class merge and before any
      ladder. **Mutation-locked** by
      `exact_span_same_class_precedence_tie_emits_family_candidate_not_merge`.
- [x] **Exact-span never calls `remove_overlaps`** — guarded by
      `overlap != Overlap::Exact` on both the `Family` and `CandidateWins` arms.
      **Mutation-locked** by `exact_span_family_tie_skips_overlap_removal` and
      `exact_span_candidate_win_skips_overlap_removal`, each of which fails when
      its own guard is dropped.
- [x] **Containment validator-preference asymmetry** — same-class containment
      still prefers the validator-backed span (`ConflictTier::Validator`) before
      the ladder, and still bypasses collision policy and anchors.
- [x] **Disjointness invariant** — enforced by the new proptest, mutation-checked.
- [x] **Multi-overlap fixed point** — `remove_overlaps` and the insertion loop
      are untouched; `multi_overlap_replacement_leaves_disjoint_set` green.
- [x] **Public API unchanged** — `resolve_candidates`,
      `resolve_candidates_with_policy`, and the `pub(crate)` anchor entry point
      keep their signatures. `Overlap` and `Arbitration` are private.
- [x] **Anti-recommendation honored** — *not* restructured into an "anchor
      pre-pass". The anchor rung still runs inside per-pair arbitration, after
      family tie and collision policy, so a family-level token can never lose to
      a rival it should have absorbed (axis-1 regression avoided).

## North star

- **Axis 1 (never leak)** — unaffected by construction and by the differential
  fuzz: no span or class outcome changes, so nothing tokenized before is left in
  the clear now.
- **Axis 2 (reversibility)** — manifest and restore path untouched;
  `prop_manifest_invariants` green.
- **Axis 4 (trust)** — the point of the change. `decided_by` now names the tier
  that actually decided, so an adopter reading an audit row for
  `AnchoredContext` is genuinely looking at a missing-anchor fallback. Commit
  `6a18c99` extends the same standard to the tests themselves: an invariant that
  survives its own mutation is not locked.

Scope is `resolver.rs` plus its tests and the CHANGELOG. S02-F2 (`min_score` /
NaN) is a later wave and is untouched.

Audit: solo scratchpad 7201 S02-F1
Resolves solo todo #2948
Locks solo todos #2963, #2964


Naoray:
Blocking review: the implementation is clean, output-identical except the intended decided_by correction, and every required local/CI gate passes. Two required mutation checks are not locked by tests:\n\n1. Family-tie-before-merge. Moving the family-tie probe below the exact same-class merge left all precedence-tie tests green. Add an exact-span, same-class, equal-precedence family case that expects the family-level candidate and fails under that mutation. Solo todo #2964.\n\n2. Exact-span no-remove_overlaps. Removing both exact-span guards left prop_resolver_invariants green at 2,048 cases. Add focused internal coverage that fails when either exact-span path invokes remove_overlaps. Solo todo #2963.\n\nIndependent verification: 5,000 old-vs-PR generated cases produced the same labels-ignored digest, while strict digests differed; reintroducing the old anchor/stale-tier behavior made both unit-level and persisted audit-row assertions fail as intended. CI is 11/11 green. No merge performed.

PR #436, opened 2026-08-17
[agent] fix(safety-net): lower mode×fallback to one total decision; fallback audit rows state what happened; one documented default

Naoray:
## Summary

S01-F1, rewritten per plan 7202 §4 item 9. Three things, in order of consequence:

1. **A leak.** Under the shipped default policy (`Resolve` + `Redact` fallback), residual PII could
   ship in the clean document with no audit row and an `Ok` return. Found while writing the
   twelve-pair table this todo asked for; not part of the original audit finding.
2. **The lowering.** `SafetyNetMode` × `SafetyNetFallback` spells twelve pairs for six behaviours.
   They are now lowered once, totally, in `SafetyNetPolicy::decision()`; every arm reads the
   decision.
3. **One documented default.** `Default` (`Resolve`/`Redact`) and the convenience `clean()`
   (`Strict`/`Redact`) disagreed. `clean()` now follows `Default`.

## 1. The leak (axis 1 + axis 2 + axis 4)

`apply_safety_net_policy`'s `Resolve` arm handed the **primary** `LeakReport` to
`apply_safety_net_fallback`, regardless of which report produced the fallback reason.

When the resolve pass *converges* and the post-resolution re-run flags a residual, that residual
lives in the **re-run** report, at post-resolve coordinates. Consequences on `main`:

| primary report | old behaviour | reachability |
|---|---|---|
| **non-empty** | redactor points at **stale pre-resolve spans**. Those spans are now inside a token resolve just minted, so it **deletes the token**, drops its manifest entry, and **leaves the residual in place** | any deterministic net — this is the broadly reachable variant |
| empty | fallback has nothing to act on: **residual ships**, **zero audit rows**, caller gets `Ok` | needs a backend whose verdict differs on byte-identical text, since a converged resolve leaves the document unchanged |

Only the `strict` fallback was safe, because it rejects the document without consulting the report.
The default `redact` fallback is the leaking one.

Reproduced on unmodified `main` before the fix — the **non-empty** variant (`Resolve`/`Redact`, a
net flagging `Dr. Schmidt` on pass 1 and `tail` on the re-run): result `Ok(())`, clean text
`" tail"`. The minted token's bytes were deleted and the residual survived. Regression:
`resolve_fallback_does_not_redact_stale_pre_resolve_spans`.

## 1b. …and the first fix was not enough (review blocker #2961)

Passing the residual report is necessary but **not sufficient**. That report can also contain
suspects lying wholly inside a live token, which the resolver classifies as *already protected*
precisely because redacting one destroys restore. `apply_safety_net_fallback` redacted every
suspect in whatever report it was given, so a follow-up mixing a genuine residual with a protected
class mismatch deleted **both**:

```
PR436_MIXED clean=" " manifest_len=0
the already-protected email token must survive fallback: left 0, right 1
```

`classify_fallback_suspects` now runs before the fallback: it emits the `Preserve` no-op audit row
for every protected suspect and hands the fallback only the actionable subset. Those `Preserve`
rows are **new**, not relocated — both producers of a fallback reason return from inside their
classification loop, so a protected suspect previously reached neither the redaction filter nor the
audit.

`Redact` *mode* is deliberately unchanged and still acts on every suspect: swallowing an
overlapping manifest entry is that mode's documented axis-2 cost, and it is exactly why only the
`Resolve` fallback — which promises to preserve what resolve protected — filters.

## 2. `SafetyNetDecision` — one total lowering

```rust
pub enum SafetyNetDecision {
    Observe { strict: bool },
    Redact,
    Resolve { on_residual: SafetyNetFallback },
}
```

`SafetyNetPolicy::decision()` is total over the closed sets and pure (a function of the two public
fields, so it cannot drift from them the way a cached field could). `apply_safety_net_policy`,
`apply_safety_net_fallback`, `run_safety_nets`, `should_skip_safety_nets`, both structured walkers,
and the CLI boundary all read it. Public `mode`/`fallback` fields unchanged.

### The twelve pairs, before and after

Behaviour is unchanged except where marked. Fixture A: the net flags an uncovered span on the first
pass. Fixture B: nothing on the first pass, a residual on the re-run.

| mode | fallback | decision | fixture A | fixture B (before → after) |
|---|---|---|---|---|
| `Strict` | `Strict`/`Tolerant`/`Redact` | `Observe { strict: true }` | report only, bytes unchanged | report only (unchanged) |
| `Tolerant` | `Strict`/`Tolerant`/`Redact` | `Observe { strict: false }` | report only, bytes unchanged | report only (unchanged) |
| `Redact` | `Strict`/`Tolerant`/`Redact` | `Redact` | suspect span deleted | no re-run, so no residual step (unchanged) |
| `Resolve` | `Strict` | `Resolve { on_residual: Strict }` | tokenized, converges | `Err(SafetyNetFallback)` + **0 → 1 audit row** |
| `Resolve` | `Tolerant` | `Resolve { on_residual: Tolerant }` | tokenized, converges | residual ships + **0 → 1 audit row** |
| `Resolve` | `Redact` | `Resolve { on_residual: Redact }` | tokenized, converges | **residual ships → residual deleted**, + **0 → 1 audit row** |

The seven pairs that differ only in an unread `fallback` are now *documented* as such on
`SafetyNetPolicy`, not silently ignored.

## 3. Audit label

`fallback_action` → `fallback_row_action`, documented as the row's claim about the suspect's bytes:
`Action::Redact` only when the span is actually deleted, `Action::Preserve` when it is left in place
(shipped under `tolerant`, rejected under `strict`). With fix 1, a `decided_by: Fallback` row also
names the residual suspect that drove it rather than a stale primary-pass suspect.

## 4. One default

`clean_with_safety_net` / `clean_with_safety_net_detect_context` used a private `Strict`/`Redact`
policy that contradicted `SafetyNetPolicy::default()`. They now use `SafetyNetPolicy::default()`
(`Resolve`/`Redact`, shipped since v0.8.1 and the CLI default).

**Behaviour change for adopters** who call those entry points *with a registered safety net*: they
now get enforcement instead of observation. Axis 1 strengthens; axis 2 is preserved (promoted spans
stay restorable). Report-only callers should pass an explicit policy to
`clean_with_safety_net_policy_detect_context`, or use `scan_safety_nets`. In-tree pipelines that
register no safety net (`gaze-document`, `gaze-proxy`'s residual validator) are unaffected —
`run_safety_nets` returns early when no net is registered.

## 5. Dead plumbing — what was kept, and why

- **`GazeLocalProtectionTraceKind::SafetyNetFallbackRedact` is reachable — kept and now tested.**
  The audit recorded it as never-constructed; on current `main` it is constructed whenever a
  fallback redaction runs with a protection trace. Not deleted; pinned by
  `fallback_redaction_is_traced_as_fallback_redact`, which is also the triple
  `clean_for_bench.rs` asserts on.
- **The `Strict`/`Redact` arms of `apply_safety_net_fallback` are not byte-identical** and were not
  merged: `Strict` returns `Err`, `Redact` mutates.
- **CLI runtime warning deleted.** `"--safety-net-fallback is ignored when --safety-net-mode is
  terminal"` is redundant now the lowering is documented. The tolerant-deprecation warning was
  narrowed to where a tolerant disposition is reachable; it no longer fires for
  `redact` + `tolerant`, where the fallback is never consulted. The `GAZE_ALLOW_TOLERANT` gate is
  untouched and still rejects a tolerant flag in any position.
- **Anti-recommendation honoured:** the fallback is *not* wired into `Redact` mode.

## 6. Docs

`docs/explanation/safety-net/safety-net-modes.md` §6 claimed a `redact` → fallback cascade the
runtime has never had; §6.2's matrix had a `redact` primary row; the TL;DR and §6.1/§6.5 repeated
the claim and warned about an unconsulted flag whose runtime warning this PR deletes. Corrected,
with an "Implemented behaviour (v0.9)" lowering table at the top of §6 that names which
subsections were rewritten, and the residual-report contract stated.

`gaze-cli/README.md`'s `--safety-net-fallback` row said the flag applies to `redact` — corrected —
and its `--safety-net-mode` row told adopters `redact` "overwrites the suspect span with a
sentinel" when the code **deletes** the bytes, so anyone grepping output for a marker finds
nothing. Also corrected. The sentinel language inside `safety-net-modes.md` is deliberately left
alone: §4.1/L21 treats the sentinel as an open design question, so changing it is a separate
decision, not a docs fix.

`safety-nets.md` now separates the *backend* observer-only contract (a net can never rewrite bytes)
from the *pipeline* policy (which may act on the report).

## 6. Also fixed — a residual the caller could not see (#2959)

The report handed back to the caller is the first pass's, so a residual found only by the re-run
was invisible in it. A boundary that decides on the report — the CLI's tolerant-mode deprecation
warning, an adopter's "did anything leak?" check — was told nothing was found while the residual
shipped under `tolerant` or was destroyed one-way under `redact`. The suspects the fallback acted
on are now merged into the returned report; a converged resolve merges nothing, so a deterministic
net that re-reports the same suspect produces no duplicates.

## Tests

- `safety_net_policy_lowering_covers_all_twelve_representable_pairs` — exhaustive, behavioural, over
  two fixtures; asserts the decision *and* the observable outcome for each pair.
- `resolve_fallback_does_not_redact_stale_pre_resolve_spans` — the broadly reachable leak variant.
- `resolve_fallback_redacts_the_residual_report_not_the_stale_primary_report` — the empty-primary
  variant.
- `resolve_fallback_redacts_the_residual_without_deleting_protected_live_tokens` — the round-2
  blocker on the **post-resolution** branch, over **both** fallback reasons (`ResidualSuspect` and
  `OverlapConflict`, which take the same unfiltered path), asserting the token survives, restore
  round-trips, and the audit carries one `Redact` row for the residual plus one `Preserve` row for
  the protected suspect.
- `first_pass_refusal_redacts_only_the_actionable_suspect_and_audits_the_protected_one` — the same
  contract on the **up-front refusal** branch, and the one that needs no stateful backend: a plain
  deterministic net reporting, on pass 1, a span inside the minted token plus a span the resolver
  cannot honor. Covers both refusal reasons (a class mismatch outside every token →
  `OverlapConflict`; an uncovered span slicing a multi-byte character → `ResidualSuspect`).
- `returned_report_surfaces_a_residual_that_only_the_re_run_found` and
  `returned_report_is_not_inflated_when_resolve_converges` — #2959 and its converse.
- `fallback_redaction_is_traced_as_fallback_redact` — trace-variant reachability.
- `clean_with_safety_net_convenience_api_uses_the_documented_default` — convenience API == `Default`.
- Two existing tests that pinned observer-only behaviour *through the convenience API* now request
  the observer policy explicitly (`observer_text_path_returns_manifest_and_report_without_mutating`,
  `byte_equal_invariance_for_leak_kinds_and_locale_skip`).

**Mutation probes.** Both fixes are load-bearing. Restoring the old report argument:

```
resolve_fallback_redacts_the_residual_report_not_the_stale_primary_report ... FAILED
  assertion failed: !clean.contains(RESIDUAL_MARKER)
safety_net_policy_lowering_covers_all_twelve_representable_pairs ... FAILED
  Resolve/Strict — fallback audit rows: left 0, right 1
```

Restoring the unfiltered suspect set (`classify_fallback_suspects` → `redaction_suspects`):

```
resolve_fallback_redacts_the_residual_without_deleting_protected_live_tokens ... FAILED
  ResidualSuspect: the protected token must survive fallback: left 0, right 1
```

Reverting **only** the first-pass branch, keeping the filter on the post-resolution one (mutation
P3 from the round-2 review — this shape previously left every test green):

```
first_pass_refusal_redacts_only_the_actionable_suspect_and_audits_the_protected_one ... FAILED
  ClassMismatchOutsideTokens: the protected token must survive the fallback: left 0, right 1
```

## Gate pinning

The four leak regressions plus the fallback-trace test are now required **by name** in
`crates/xtask/src/safety_net_sanity.rs` (18 pinned tests), so a refactor that drops one turns the
gate red rather than silently retiring the proof of a shipped leak. Mutation-checked — renaming one
locally:

```
safety_net_sanity: checking 18 behavioral tests across 4 batched cargo test invocations
Error: safety_net_sanity: missing required test
  `first_pass_refusal_redacts_only_the_actionable_suspect_and_audits_the_protected_one`
  in core safety-net manifest and structured behavior
```

## Gates

`cargo fmt --all -- --check`, `cargo clippy --workspace --all-features --all-targets -D warnings`,
`cargo test --workspace --all-features --no-fail-fast` (131 binaries at `57feb33`),
`cargo test -p gaze-pii --all-features` and `cargo run -p xtask -- safety-net-sanity` at the fix
commit.

## North star

Axis 1 (a real leak closed, fail-closed preserved, no fallback wired into `redact`), axis 2
(resolve stays reversible; the fallback no longer mangles minted tokens), axis 4 (the audit row
states what happened; a 12-state type with 7 ignored states is now a 6-state one), axis 5 (one
documented default). No axis weakened; the `clean()` default change is the one adopter-visible
behaviour change and is called out in the CHANGELOG.

Audit: solo scratchpad 7201 S01-F1 (rewritten per plan 7202 §4 item 9). Review: scratchpads 7330,
7334.

Resolves solo todo #2949. Refs #2956, #2959, #2960, #2961, #2970.


Naoray:
## Review record — three rounds on a leak path

This PR fixes a **shipped defect**: the safety-net `Resolve` fallback acted on the primary `LeakReport` rather than the one that produced the fallback reason. Two reviewers ran in parallel each round (code + semantics), and the fix took three rounds. Recording the path here because this repo is public and a leak-path fix that needed three rounds should show its work.

### The original defect

With a **non-empty** primary report — the broadly reachable case, since any deterministic safety net produces one — the redactor was pointed at stale pre-resolve spans. Those spans had since become part of a token the resolve pass minted, so the fallback deleted the token, dropped its manifest entry, and left the actual residual in the document: an axis-1 leak and an axis-2 restore break in one operation, returned as `Ok`.

With an **empty** primary report the fallback had nothing to act on at all, so the residual shipped and no fallback audit row was written. (Reaching that shape additionally requires a backend whose verdict differs across byte-identical text, since a converged resolve leaves the document unchanged.)

Only the `strict` fallback was safe, because it rejects the document without consulting the report.

### Round 1 → blocker

The first fix handed the fallback the *complete* post-resolution report. Correct for locating the residual, but not sufficient: `redaction_suspects` does not filter suspects lying inside live tokens, and `post_resolution_fallback_reason` returns from *inside* its classification loop, above its own `Preserve` logging. So a mixed report — one already-protected suspect plus one genuine residual — deleted the minted token **and** left the protected suspect with no audit row at all. An axis-2 regression introduced by the fix itself.

### Round 2 → blocker

The rework was correct, and went further: the same flaw existed on the **first-pass refusal** path, where `resolve_safety_net_suspects` bails before mutating anything, and it was fixed there too.

But that second branch was **untested**. Reverting only it left 29/29 safety-net and 185/185 lib tests green, with all four pinned regressions passing. It is also the *more* reachable branch: the post-resolution fixtures need a stateful backend, whereas the first-pass refusal needs only one ordinary deterministic net flagging, on a single pass, one span inside an already-minted token plus one span that cannot be honored.

A fix exercised by only one fixture shape is exactly what produced the round-1 blocker, so this was blocked rather than waved through.

### Round 3 → clean

`first_pass_refusal_redacts_only_the_actionable_suspect_and_audits_the_protected_one` closes it, using a net with no call counter and covering both refusal reasons: a class mismatch outside every token (`OverlapConflict`) and a span slicing a multi-byte character (`ResidualSuspect`).

### Mutation probes

Every guarantee below was shown load-bearing by reverting it and confirming the suite goes red — run by the reviewer independently, not accepted from the author.

| probe | mutation | outcome |
|---|---|---|
| P2 | live-token filter reverted, both branches | mixed test **red** — protected token deleted, `manifest.len()` 0 vs 1 |
| P2b | P2, `OverlapConflict` case ordered first | **red** — both fallback reasons fail independently |
| P3 | first-pass branch only reverted | round 2: all green (the blocker) → round 3: **red** |
| P3b | P3, `NonCharBoundary` case ordered first | **red** — both refusal reasons fail independently |
| P4 | one pinned regression renamed, allowlist untouched | gate bails, **true exit 1** |
| P5 | fallback reads the primary report again | **6 tests red** — the original fix stays pinned |
| P6 | filter kept, `Preserve` logging dropped | **red** — the audit assertion is load-bearing, not decorative |

P6 and P4 are the two that matter most for trust. P6 proves the protected suspect is genuinely *audited*, not merely spared from redaction — filtering it out of the redaction set is only half the fix, and a test that checked only the bytes would have missed the missing row. P4 checks the gate's **exit code**, not just its error message, because a gate that prints an error and exits 0 is the failure mode that pinning exists to prevent.

### Regressions pinned by name in `safety-net-sanity`

- `resolve_fallback_redacts_the_residual_report_not_the_stale_primary_report`
- `resolve_fallback_redacts_the_residual_without_deleting_protected_live_tokens` (both fallback reasons)
- `resolve_fallback_does_not_redact_stale_pre_resolve_spans`
- `first_pass_refusal_redacts_only_the_actionable_suspect_and_audits_the_protected_one` (both refusal reasons)
- `fallback_redaction_is_traced_as_fallback_redact`

Pinning by name means a refactor that quietly drops or renames one turns the gate **red** instead of silently retiring the proof. The gate now checks 18 behavioural tests across four suites.

### Adopter-visible changes

- **`Pipeline::clean_with_safety_net` / `clean_with_safety_net_detect_context`** — the policy-less convenience entry points — now use `SafetyNetPolicy::default()` (`Resolve` + `Redact`) instead of a private `Strict` + `Redact` that contradicted it. **Callers with a registered safety net now get enforcement instead of observation:** suspects are tokenized reversibly and any residual takes the `Redact` fallback. For the previous behaviour, pass an explicit `SafetyNetMode::Strict` policy to `clean_with_safety_net_policy_detect_context`, or use `Pipeline::scan_safety_nets` for a report-only pass. Every in-tree production caller registers no safety net, so all are unaffected.
- **A residual found only by the post-resolution re-run now appears in the returned `LeakReport`.** Previously a boundary deciding on that report was told nothing was found while the residual shipped under `tolerant` or was destroyed one-way under `redact`. A converged resolve merges nothing, so deterministic nets do not produce duplicates.
- **`gaze clean`** no longer warns that `--safety-net-fallback` is ignored for terminal modes, and the tolerant-deprecation warning now fires only where a tolerant disposition is actually reachable.
- **`--safety-net-mode redact` deletes the suspect bytes outright** — they are removed, not replaced with a sentinel string, so there is no marker to grep for. The CLI reference previously said otherwise and is corrected here; the wider modes-doc rewrite is tracked separately.

Two harmless behaviour notes for anyone bisecting later: the `ClassMismatch` stderr warning can now fire on runs where it previously did not, since an actionable class-mismatch residual reaches the returned report; and `LeakReport::extend` clears `replay_hash`, which is inert today because nothing sets it.

PR #439, opened 2026-08-17
[agent] fix(mcp-tools): classify tokens from the session snapshot instead of two shadow token-grammar parsers

Naoray:
## Summary

- classify restore responses from the matching `SessionSnapshotEntry` instead of parsing token text
- build export entries directly from one `Session::snapshot_entries()` result
- delete both duplicated token-grammar parsers and their internal error types
- preserve the existing response wire shapes and `ToolError::NotFound` fail-closed behavior

## Safety and compatibility

- `Name` and `Custom("order_id")` now round-trip through both restore and export with class labels asserted against `PiiClass::class_name()`
- the new class tests also passed on unmodified `main`; this fixes architectural drift risk rather than a currently reproduced label mismatch
- no new core accessor or wire field was introduced
- reliability, reversibility, agentic fit, trust, and adopter ergonomics are preserved or strengthened by making the manifest the class authority

## Performance tradeoff

- restore now clones the session snapshot and performs an O(n) lookup; export already processes the full inventory and now does so from one O(n) snapshot clone
- this bounded operator-tier cost is accepted to remove shadow grammar and keep classification manifest-authoritative; correctness axes 1-4 take priority

## Verification

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-features --all-targets -- -D warnings`
- `cargo test -p gaze-mcp-core --all-features`
- `cargo run -p xtask -- dylint-gate` (18 UI fixture shapes passed; cargo-dylint remains CI-scheduled per todo #1870)
- `cargo test --workspace --all-features --no-fail-fast`

All gates ran with `cargo`, `rustc`, and `clippy-driver` pinned to Rust 1.96.0 in an isolated target directory.

Audit: solo scratchpad 7201 S13-F1

Resolves solo todo #2955

### DIFF
--- a/crates/gaze-mcp-core/src/tools/export.rs
+++ b/crates/gaze-mcp-core/src/tools/export.rs
@@ -46,20 +46,16 @@ impl Tool for ExportSessionTokensTool {
     async fn invoke(&self, ctx: &ToolCtx<'_>) -> Result<ToolResponse, ToolError> {
         let session = ctx.resources().session();
         let mut entries = session
-            .tokens()
+            .snapshot_entries()
             .into_iter()
-            .map(|token| {
-                let raw = session
-                    .restore_strict(&token)
-                    .map_err(ToolError::internal)?;
-                let class = classify_token(&token)?;
-                Ok(json!({
-                    "token": token,
-                    "raw": raw,
-                    "class": class,
-                }))
+            .map(|entry| {
+                json!({
+                    "token": entry.token,
+                    "raw": entry.raw,
+                    "class": entry.class.class_name(),
+                })
             })
-            .collect::<Result<Vec<_>, ToolError>>()?;
+            .collect::<Vec<_>>();
         entries.sort_by(|left, right| left["token"].as_str().cmp(&right["token"].as_str()));
         let count = entries.len();
         Ok(ToolResponse::json(json!({
@@ -69,45 +65,6 @@ impl Tool for ExportSessionTokensTool {
         })))
     }
 }
-
-fn classify_token(token: &str) -> Result<String, ToolError> {
-    if token.starts_with("email") && token.ends_with("@gaze-fake.invalid") {
-        return Ok("Email".to_string());
-    }
-    let core = token
-        .strip_prefix('<')
-        .and_then(|value| value.strip_suffix('>'))
-        .unwrap_or(token);
-    let after_session = if core.len() > 9
-        && core.as_bytes()[8] == b':'
-        && core[..8].bytes().all(|byte| byte.is_ascii_hexdigit())
-    {
-        &core[9..]
-    } else {
-        core
-    };
-    let class_part = after_session
-        .rsplit_once('_')
-        .map(|(class, _)| class)
-        .ok_or_else(|| ToolError::internal(TokenClassError(token.to_string())))?;
-    if let Some(custom) = class_part
-        .strip_prefix("Custom:")
-        .or_else(|| class_part.strip_prefix("custom:"))
-    {
-        return Ok(format!("Custom:{custom}"));
-    }
-    Ok(match class_part {
-        "email" => "Email".to_string(),
-        "name" => "Name".to_string(),
-        "location" => "Location".to_string(),
-        "organization" => "Organization".to_string(),
-        class => class.to_string(),
-    })
-}
-
-#[derive(Debug, thiserror::Error)]
-#[error("could not classify token `{0}`")]
-struct TokenClassError(String);
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


Naoray:
## Review: BLOCKED on one CI-red gate — the production change itself is clean

Reviewed at head `37a0b26` in an isolated worktree, rustc pinned to 1.96.0. **Not merged.** One blocker, two nits. The blocker is in the new **test fixtures**, not in the logic — a ~4-line fix.

### BLOCKER — `xtask gates` red at `no-tenant-knowledge`

[Job 95521475345](https://github.com/CertaMesh/gaze/actions/runs/32073456531/job/95521475345), step 11 `no-tenant-knowledge => failure` (steps 12–22 skipped as a consequence). Reproduced locally:

```
$ cargo run -q -p xtask -- no-tenant-knowledge
Error: NoTenantKnowledgeDenylistHit
crates/gaze-mcp-core/src/tools/export.rs:166 matched order_id: (gaze::PiiClass::Custom("order_id".to_string()), "ORDER-0001"),
crates/gaze-mcp-core/src/tools/restore.rs:180 matched order_id: (gaze::PiiClass::Custom("order_id".to_string()), "ORDER-0001"),
```

The new fixtures hardcode the denylisted `order_id` inside production-crate `src/`:

- DENYLIST (`crates/xtask/src/no_tenant_knowledge.rs`) = `order_id`, `Order_42`, `Song_42`, `User_7`.
- `repo::production_source_dirs()` covers **every** workspace member's `src/` except `EXEMPT_CRATES` (xtask), so `gaze-mcp-core/src/` is in scope.
- `repo::tenant_knowledge_file()` accepts every `.rs` file and deliberately does **not** exempt inline `#[cfg(test)]` modules — contrast `repo::fixture_citation_file()`, which does exempt `tests`/`test_support` stems.
- No escape hatch: `// allow(tenant-fixture)` **hard-fails** in production scope.

**Fix (preferred):** rename the fixture class to a generic non-denylisted name, e.g. `PiiClass::Custom("case_ref")` with raw `"CASE-0001"`. The class *name* is incidental — what the test asserts is that a Custom class round-trips. Alternatively move both tests to `crates/gaze-mcp-core/tests/` (out of scope), matching `crates/gaze/tests/token_shape_shadow_guard.rs`, which legitimately uses `custom:order_id`.

Worth noting this was an easy trap to fall into: `crates/xtask/README.md:46` still documents the pre-#434 crate list (`gaze`, `gaze-types`, `gaze-recognizers`, `gaze-assembly`, `gaze-cli`), which makes `gaze-mcp-core` look out of scope. That list went stale when #434 replaced the hardcoded `PRODUCTION_CRATES` lists with a scope derived from workspace members — which is precisely why this crate is scanned now and was not before. Tracked separately as solo todo #2975; no action needed from you on that one. Adding this gate to the PR's verification list would have caught it locally.

### Everything else verified green

Local gates, rustc 1.96.0, isolated target dir:

```
cargo fmt --all -- --check                                            -> OK
cargo clippy --workspace --all-features --all-targets -- -D warnings  -> clean
cargo test -p gaze-mcp-core --all-features                            -> 38 pass, 0 failed
cargo test -p gaze-cli --all-features                                 -> 193 pass, 0 failed
```

**Wire-shape invariance — no adopter-visible change.** I reconstructed the deleted `classify_token` verbatim and compared it to `PiiClass::class_name()` across every class shape × both token grammars: all 4 `builtin_variants()` plus 9 custom shapes (`::custom("order_id")`, literal, no-underscore, trailing-digit, single-char, empty `Custom("")`, `family("iban_like")`, `Custom("email")`, `Custom("name")`), each tokenized through both `Session::tokenize` and `format_preserving_fake`. Result: **26 comparisons, 0 mismatches.** The equivalence has a clean root cause — the token grammar embeds `class_name()` verbatim, so the old parser was reading back the very string `class_name()` produces.

**Fail-closed preserved.** Old path: `restore()` → `None` → `ToolError::NotFound(format!("token {token} not in session"))` before classify ever ran. New path: snapshot `.find()` → `None` → the same `ok_or_else(NotFound)`, identical message. Same variant, same string, no new silent-`Ok`.

**No widened exposure.** `tokens()` reads `value_by_token.keys()`; `snapshot_entries_from_state()` reads `token_by_value` — two directions of one map whose bijectivity is enforced on import. Same token set. Export also loses its `restore_strict` error arm, but that inconsistency is now impossible by construction: raw and class come from one entry and can no longer disagree. That is the real value of this change.

### On the tests passing against the old parser

Your disclosure — *"the new class tests also passed on unmodified `main`"* — is accurate, and I ran the reconstruction to confirm it (hybrid = old `classify_token` + new tests → 28 pass, 0 fail). I am **not** treating it as a blocker, because the invariance result above shows the two implementations agree on every reachable input, so no correct-value assertion could discriminate them. Demanding the tests go red would be demanding proof of a behavior change, which is the opposite of what a dedup refactor should do.

I did verify the tests aren't vacuous, by mutating `entry.class.class_name()` → `to_canonical_str()`:

```
restore_uses_manifest_class_for_builtin_and_custom_tokens ... FAILED   left: "name"   right: "Name"
export_uses_manifest_class_for_builtin_and_custom_tokens  ... FAILED   left: "name"   right: "Name"
export_emits_session_token_inventory                      ... FAILED   left: "email"  right: "Email"
restore_emits_raw_payload_for_known_operator_token        ... FAILED   left: "email"  right: "Email"
```

They are live invariance guards. Thanks for stating the situation plainly in the PR body rather than implying a reproduced mismatch.

### NIT (non-blocking, pre-existing — filed, no action needed here)

Every restore/export test, new and pre-existing, builds tokens via `Session::tokenize` (`<{hex}:{ClassName}_{i}>`). The Pipeline emits `format_preserving_fake` shapes instead (`email{i}.{hex}@gaze-fake.invalid`, `{hex}:{lowercase}_{i}`, `{hex}:custom:{name}_{i}`), and `Location`/`Organization` labels are unpinned. This matters *less* after your change, since classification became grammar-independent — before it, those were exactly the branches the deleted parsers had and no test exercised.

### To unblock

Rename the two fixture classes off `order_id`, re-run `cargo run -p xtask -- no-tenant-knowledge` and the CI-canonical clippy, push. On that basis this is a merge: the change is correct, invariant, fail-closed, and removes a genuine single-owner violation of the token-grammar contract.

PR #573, opened 2026-09-14
fix(gaze-mcp-bridge): evict only persisted inactive sessions under the max_sessions cap

### LINKED ISSUE #490: [Detail Bug] MCP bridge sessions grow unbounded in memory for unique session IDs
# Detail Bug Report

https://app.detail.dev/org_cb6b545f-80ba-408b-91dd-c38dbcbaec9c/bugs/bug_16ce73f6-7215-4272-87fe-442d6a828a2b

Introduced in [#330](https://github.com/CertaMesh/gaze/pull/330) by @Naoray on Jun 20, 2026

# Summary
- **Context**: `BridgeSessionStore` in `crates/gaze-mcp-bridge/src/session_store.rs` is the in-process cache of live `gaze::Session` objects for the MCP bridge, keyed by the caller-supplied external session id. The store is a public type in a published library crate (`pub use session_store::{BridgeSessionStore, SessionStoreMode}` in `lib.rs`).
- **Bug**: The store inserts into its `sessions` and `file_locks` `HashMap`s on every cache miss and never removes, evicts, or bounds either map, so memory grows monotonically with the number of distinct session ids seen over the lifetime of the process. There is no capacity cap, no TTL, no eviction sweep, no `Drop` cleanup, and no configuration field (`SessionCfg` exposes only `mode`, `dir`, `key_env`). The store's fields are private and it exposes only `from_config`, `ephemeral`, `get`, and `persist` — no eviction API — so the host process cannot reclaim cache entries even if it manages caller lifecycle.
- **Actual vs. expected**: A long-running bridge process should retain only a bounded working set of sessions. The store has no bound of any kind — neither configurable via `SessionCfg` nor hardcoded.
- **Impact**: The mechanism — a monotonic insert-only HashMap with no eviction path — is a resource leak by construction; time-to-exhaustion depends on deployment and cannot be measured in this environment.

# Code with Bug
`crates/gaze-mcp-bridge/src/session_store.rs`:

```rust
pub struct BridgeSessionStore {
    mode: SessionStoreMode,
    sessions: Mutex<HashMap<String, SharedSession>>,        // <-- BUG 🔴 unbounded insert-only cache
    file_locks: Mutex<HashMap<String, Arc<Mutex<()>>>>,    // <-- BUG 🔴 unbounded insert-only lock registry
}
```

```rust
pub async fn get(&self, validated_session_id: &str) -> BridgeResult<SharedSession> {
    let mut sessions = self.sessions.lock().await;
    if let Some(session) = sessions.get(validated_session_id) {
        return Ok(Arc::clone(session));
    }
    let session = match &self.mode { /* Ephemeral | File load */ };
    let shared = Arc::new(Mutex::new(session));
    sessions.insert(validated_session_id.to_string(), Arc::clone(&shared)); // <-- BUG 🔴 inserts on miss, never evicted/removed
    Ok(shared)
}

async fn file_lock(&self, validated_session_id: &str) -> Arc<Mutex<()>> {
    let mut locks = self.file_locks.lock().await;
    Arc::clone(
        locks
            .entry(validated_session_id.to_string())      // <-- BUG 🔴 inserts on miss, never evicted/removed
            .or_insert_with(|| Arc::new(Mutex::new(()))),
    )
}
```

# Explanation
- `BridgeSessionStore::get()` caches a `SharedSession` per `validated_session_id` and always inserts on cache miss.
- There is no code path or API that ever removes entries from `sessions` or `file_locks`, and no configuration supports a cap/TTL.
- Session IDs are caller-controlled (`tools/call` `_session_id`) and validation enforces only format/entropy-floor checks; it does not restrict cardinality (e.g., “new ULID per call” is policy-valid). In long-lived processes handling many unique session IDs, memory usage grows monotonically.
- The bug is reachable when embedding the bridge library in a long-lived host with permissive auth; the shipped `gaze mcp bridge` CLI uses `DenyAllAuthHook` by default, so it typically fails before reaching `session_store.get`.

## Codebase Inconsistency
`gaze daemon` (a long-lived server in the same repo) manages the same `gaze::Session` type with a session cap and LRU eviction (e.g., `--session-cap` default 1000 and `while self.sessions.len() >= self.session_cap { ... self.sessions.remove(&evicted) ... }`). This provides an existing eviction pattern that the bridge store currently lacks.

# Recommended Fix
- Apply bounding/eviction **only for `SessionStoreMode::File`** (evicted sessions can be reloaded from disk; eviction in Ephemeral mode would permanently lose tokenization state).
- Add `max_sessions` to `SessionCfg` (with `#[serde(default)]`; suggested default `1000`) and implement an LRU-based cap in `BridgeSessionStore::get()` for File mode (track access order and evict least-recently-used sessions when the cap is exceeded).
- Note: `file_locks` remains unbounded under this fix; either document it as a secondary leak or add a safe eviction scheme later.

# History
This bug was introduced in commit 3e9ea44. The commit added the entire `gaze-mcp-bridge` crate (17 files, ~3000 LOC, squashed PR #330) and authored `session_store.rs` from scratch with the unbounded insert-only `sessions` and `file_locks` HashMaps already in their final buggy form — no later commit ever modified the file, so `git blame` on every relevant line points here. The bug slipped in because the author modeled only the cache-miss → insert path and omitted the cache-full → evict half of the pattern, even though the same codebase already implemented LRU + `session_cap` eviction for `gaze::Session` in `crates/gaze-cli/src/commands/daemon.rs` (visible at this same commit: `daemon.rs:134` declares `lru: VecDeque<String>`, `daemon.rs:296` calls `evict_idle_sessions`) — the proven pattern was available locally but not carried over to the bridge store.

Naoray:
## What & why

Bound bridge session caching with `session.max_sessions` (default 1,000; zero rejected). Ephemeral mode rejects admission at capacity. File mode persists the least-recently-used inactive candidate before eviction, preserving the canonical manifest after failed persistence or caller cancellation. If persistence fails, admission returns the typed `BridgeError::SessionStore` and retains the cached session. If every cached session is active, admission returns `BridgeError::LimitExceeded`.

One `SessionCache` mutex owns entries and recency. Cache replacement happens only after persistence and loading succeed, without an intervening await. Always persisting the current candidate avoids a dirty flag that could miss public `SharedSession` mutations. The existing `BridgeHost::dispatch_inner` persistence-error path needs no change.

Supersedes #522
Resolves #490
Resolves solo todo #3527

## Blocker resolution

> “evict only persisted, inactive entries.”

`crates/gaze-mcp-bridge/src/session_store.rs::get` retains the active-Arc guard and calls `persist` before removing the inactive entry. The reviewer's unchanged `r2_failed_persist_must_not_allow_dirty_session_eviction` now passes. Both `w1_file_cap_retains_active_session_identity` and `r2_original_w1_file_cap_retains_active_session_identity` remain green.

> “persist the inactive candidate before removing it and reject admission on persistence failure.”

`eviction_persists_unpersisted_candidate_before_restore_round_trip` verifies a fresh store can restore the eviction-written encrypted snapshot. `eviction_rejects_admission_when_candidate_persistence_fails` replaces the directory with a file, checks the typed persistence failure, retained manifest and cap, then restores the directory and verifies successful retry and restoration.

> “Cover failed persistence and cancellation/unpersisted release, retaining the active-Arc regression.”

`eviction_preserves_unpersisted_state_after_caller_cancellation` uses a oneshot rendezvous and abort/join to release a mutated session without persistence, then checks restoration after eviction. No sleeps or permission-dependent faults. The unpersisted-release and reviewer failure cases complement it.

> “one cache state with entries, recency, and persisted/dirty lifecycle (or eviction-time persistence).”

Implemented `SessionCache` plus eviction-time persistence. The LRU test now directly checks object lifetime rather than assuming eviction loses unpersisted state. Its environment key is independent of concurrently running file-store tests.

- Verdict: implement.
- Opportunity: one cache state, with durability established at eviction.
- Why: remove mirrored mutex ownership and insert/undo eviction loops; encode inactivity and successful persistence before losing the canonical in-memory manifest.
- Scope: session-store admission, config documentation, and contract tests; reviewed immediate callers in `bridge.rs`, with no caller changes needed.
- Validation: baseline reviewer proof, source-only mutation testing, crate tests, and workspace gates below.

Axes: reliability, reversibility, and deterministic ownership improve; agent workflows retain token mappings after cancellation. Adopter ergonomics gain a default cache cap. Tradeoff: file admission at capacity adds persistence I/O while holding the cache lock and may reject requests during filesystem faults. Correctness takes priority over throughput. The cap bounds cached sessions only; the separately retained per-ID `file_locks` map is still unbounded. No visible output changes.

## Local gates

All Cargo commands used the required `g.sh <worktree> cargo ...` wrapper with Rust/Rustdoc 1.96.0 and a worktree-local target directory. Builds ran sequentially with `-j 4`.

Baseline, before production changes:
- `cargo test -p gaze-mcp-bridge --all-features -j 4 --test contract r2_`: exit 101, one pass and the expected dirty-eviction failure.
- `cargo test -p gaze-mcp-bridge --all-features -j 4 --test contract w1_file_cap_retains_active_session_identity`: exit 0, two passes.

Mutation proof after signed commit `e41ecb4`: temporarily restored only `src/session_store.rs` from its parent, retaining all tests. `cargo test -p gaze-mcp-bridge --all-features -j 4 --test contract eviction_`: exit 101, all three new tests failed. Restored the committed source before final gates. This uses path-scoped restore rather than disturbing any unrelated stash.

Initial workspace run: exit 101 in the unrelated xtask self-test `bundle_tokenization_drift_gate_fails_when_enabled_recognizer_id_drifts`: its altered fixture incorrectly passed. Serial retry (`CARGO_BUILD_JOBS=4 RUST_TEST_THREADS=1 cargo test -p xtask --all-features -j 4 --test bundle_tokenization_drift`) also exited 101: the altered fixture was correctly rejected, then the two baseline tests tried to read a deleted fixture path. Both outcomes identify inherited absolute `CARGO_TARGET_DIR` reuse across disposable workspaces. No xtask source was changed. Follow-up runs clear only that inherited variable after entering the required wrapper, explicitly set the parent `--target-dir target`, and let child workspaces use their own local targets; the pinned Rust/Rustdoc bindings remain unchanged.

The first aggregate `ci-feature-matrix` attempt was intentionally interrupted during its repeated Clippy stage to address that fixture-target problem before its nested full-suite runs.

Isolated-target drift retry: `CARGO_BUILD_JOBS=4 g.sh <worktree> env -u CARGO_TARGET_DIR cargo test -p xtask --all-features -j 4 --target-dir target --test bundle_tokenization_drift`: exit 0, all three tests passed. This establishes the target-directory workaround without changing the gate or weakening its assertions.

Final results:
- `cargo fmt --all -- --check`: exit 0.
- `cargo clippy --workspace --all-features --all-targets -j 4 -- -D warnings`: exit 0.
- `cargo test -p gaze-mcp-bridge --all-features -j 4`: exit 0; one unit test and 43 contract tests passed, one existing contract test ignored. Includes all three new regressions and both active-identity tests.
- `cargo test --workspace --all-features`: exit 0 in the corrected aggregate run; 1,754 passed, zero failed, 15 ignored across 145 suites. Nested builds used `CARGO_BUILD_JOBS=4`.
- `cargo deny check`: exit 0.
- `cargo run -p xtask -j 4 -- symmetric-potemkin`: exit 0.
- `cargo run -p xtask -j 4 -- class-map-override-safety`: exit 0.
- `cargo run -p xtask -j 4 -- recognizer-composition-validator`: exit 0.
- `cargo run -p xtask -j 4 -- no-tenant-knowledge`: exit 0.
- `cargo run -p xtask -j 4 -- bundle-tokenization-drift`: exit 0.
- `cargo run -p xtask -j 4 -- family-policy-table-coherence`: exit 0.
- `cargo run -p xtask -j 4 -- locale-cue-bundle-coherence`: exit 0.
- `cargo run -p xtask -j 4 -- fixture-citation-lint`: exit 0.
- `cargo run -p xtask -j 4 -- trybuild-fixture-hygiene`: exit 0.
- `cargo run -p xtask -j 4 -- cargo-metadata-audit-isolation`: exit 0.
- `cargo run -p xtask -j 4 -- readme-version-check`: exit 0.
- `cargo run -p xtask -j 4 -- safety-net-sanity`: exit 0.
- `CARGO_BUILD_JOBS=4 g.sh <worktree> env -u CARGO_TARGET_DIR cargo run -p xtask -j 4 --target-dir target -- ci-feature-matrix`: exit 0. All 30 aggregate commands passed, including both complete all-features and default-feature workspace suites.

https://claude.ai/code/session_01MLYvxfr2LeFKAtQ81wEbmn


Naoray:
Merge gate: BLOCKED at PR head e41ecb40e6005b1787f058934de16e64a89c970f. The original durability blocker is fixed, but public Weak revival can still split the canonical manifest. In session_store.rs:106, strong_count == 1 is checked before awaited persistence; a caller can upgrade an existing Weak while persistence is pending, and get() still removes that now-active candidate at line 135. Holding the cache mutex does not prevent Weak::upgrade(). A later lookup loads a second Session and loses mappings created through the revived handle.

Signed, DCO-compliant evidence: [8594440990e3c4229f582b8642d765f5a389b84e](https://github.com/CertaMesh/gaze/commit/8594440990e3c4229f582b8642d765f5a389b84e), branch agent/review-573-evidence, not a merge candidate. Test `r3_weak_upgrade_during_eviction_preserves_canonical_manifest` polls cap-1 admission into pending persistence, upgrades A's Weak, finishes B admission, mutates the revived A, and reacquires A. Exit 101: restore returns None instead of Some("alice@example.invalid"). This also fails when run alongside the three new eviction tests, all of which pass.

Independent original proofs: `r2_failed_persist_must_not_allow_dirty_session_eviction`, `w1_file_cap_retains_active_session_identity`, and `r2_original_w1_file_cap_retains_active_session_identity` each exit 0. Source-only mutation removed just the candidate lock/persist statements; the original r2 proof exits 101 with None instead of the original mapping. Committed production source restored afterward. After restoring the source, the original r2 proof exits 0 again. Each of the three new eviction/persistence tests was also rerun individually with `--exact`, exit 0 each.

Final local gates, on the isolated evidence checkout (PR production source unchanged, plus the added regression):
- `cargo fmt --all -- --check`: exit 0. An earlier check identified only formatting in the newly authored proof; that was corrected before the signed evidence commit.
- `cargo clippy --workspace --all-features --all-targets -j 4 --target-dir target -- -D warnings`: exit 0.
- `cargo test -p gaze-mcp-bridge --all-features -j 4`: exit 101, solely the added `r3_weak_upgrade_during_eviction_preserves_canonical_manifest`; 43 existing contract tests pass, one existing test is ignored, and the unit test passes.
- `cargo test --workspace --all-features -j 4 --target-dir target`: exit 101 at the same regression, with the same None-versus-original-mapping failure. Cargo stops there, so later workspace suites did not execute; this is not a green full-suite claim.

Every Cargo command used `g.sh <worktree>` with pinned Rust/Rustdoc 1.96.0 and sequential builds. For Clippy/workspace gates, `env -u CARGO_TARGET_DIR CARGO_BUILD_JOBS=4` after the wrapper plus explicit `--target-dir target` retains the documented isolated-target workaround for child fixtures. The original named proofs and mutation used `cargo test -p gaze-mcp-bridge --all-features -j 4 --test contract <test-name> -- --exact`.

Both PR commits, 471e12ac and e41ecb40, have SSH signature blocks, verify as G, and carry matching Signed-off-by trailers. Evidence commit 8594440 does too. Latest `gh pr checks 573`: exit 8, tests and xtask gates still pending; DCO, MSRV, docs, CodeQL/analyses and Python checks pass. Pending CI is not treated as a failure; the independently reproduced correctness defect is the merge blocker.

Required fix: establish exclusive ownership against both existing strong borrowers and weak revival throughout eviction, or use a session lease whose acquisition participates in cache ownership. Rechecking strong_count once before removal still leaves a concurrent upgrade window.

Structure/data-model review:
- Verdict: recommend.
- Opportunity: exclusive eviction ownership that accounts for public Weak handles, while retaining the unified cache.
- Why: entries and recency now have one owner, but the public Arc still permits an active session to escape the eviction invariant.
- Scope: session_store.rs candidate eligibility/ownership plus this regression; fix before merge. Immediate caller dispatch_inner retains its strong Arc correctly. No broader refactor requested.
- Validation: original proofs GREEN, core-hunk mutation RED, independent Weak-revival proof RED. Changes remain inside gaze-mcp-bridge; the separate unbounded file_locks limitation remains documented. No visible output change.

Reversibility and deterministic canonical ownership remain violated. No merge, no closure of #522, and todo #3527 remains open.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01MLYvxfr2LeFKAtQ81wEbmn

PR #596, opened 2026-09-15
Protect admitted residual raw coverage missed by whole-candidate arbitration

Naoray:
## What this changes

Whole-candidate arbitration picks one winner per overlap and emits it. Every other byte the losing originals actually covered is dropped on the floor and ships in the clear. On the pair fixture the originals cover `0..21`, the winner covers `0..15`, and `15..21` leaves the pipeline as raw text.

This PR adds the engine that protects that remainder. For a component where every member's effective class is known to tokenize, it computes `R = U - W` — the admitted original raw union minus the frozen selected raw geometry — and emits each residual cell as an ordinary reversible replacement, carrying indexed parent evidence on the B0 occurrence ledger.

**This PR includes the B0 commits** (`3036954`..`9fc6f7e`, the occurrence ledger). B0 is also open as its own PR; both are based on `main` because stacked bases get no CI in this repo. Review B0 first.

Residual coverage is **off by default** in this PR (`Pipeline::residual_coverage`, set only by tests). That is deliberate staging, not the finished state: the consumer migration and the default flip are the follow-up PR. **This PR on its own protects no additional byte in production.**

## How it works

- **Closed rule preview.** `RuleEntry` captures an immutable description at registration for exactly the three built-in rule types, by `Any` downcast before erasure. Wrappers and custom rules stay Unknown. First known match wins, known non-match continues, Unknown before a possible match stops the preview, and a fully known no-match means Preserve. No custom callback, no new `Rule` bound, no speculative action execution, no cached dynamic result.
- **Shared effective view.** Eligibility uses the resolver's existing pure effective-view and anchor-fallback helper, the real class, field context, structural membership and the actual selected result. An actual `CollisionPolicy` bypass stays bypassed; no `CollisionPolicy` is fabricated on a suppressed original.
- **Raw endpoint planner.** Strict half-open connected components; touching does not couple. `U` is admitted original evidence plus supported selected structural geometry, never an invented hull. Cells coalesce only on an exact match of parent set, representative, class, family and purpose. UTF-8, non-empty, subset, disjoint and complete-union checks all run before any residual effect.
- **Unchanged effect schedule.** Residual effects follow the existing ones: allocation, then audit, then checked replacement and trace. Earlier live state and staged audit rows are never rolled back. A preview that disagrees with the runtime action is treated as an implementation defect that suppresses the residual — never as permission to override the real action.

## Five-field cleanup verdict

1. **Verdict:** `implement` for one extraction, done here; `recommend` for the rest.
2. **Opportunity:** `resolver::candidate_order`, extracting the legacy stable candidate sort key out of `CandidatePool::new`.
3. **Why:** the residual ledger has to verify that the evidence order it was handed is the real one. Before the extraction the only copy of that comparator was inline in a constructor, so the validator would have had to re-spell it — two copies of a subtle six-level ordering with input-order ties, drifting independently. One named function removes the duplicate rule and lets `residual::validate` compare against the actual key. Byte-identical comparator; no ordering change.
4. **Scope:** 8 files inside `crates/gaze`, +2086/-64 over B0. The larger structural opportunity here is the residual planner's `I` incidence term (below) — that one should wait for measurement, not be pre-optimized.
5. **Validation:** the three gates below, plus 21 residual falsification tests and a mutation probe on each new assertion.

## North-star axes

1. **Reliability (never leak).** The point of the PR. It closes a systematic gap where arbitration itself created uncovered bytes. It is conservative on purpose: a component is admitted only when every member is known to tokenize, so it never widens coverage by guessing. It is also off by default here, so the reliability gain is not realized until the follow-up.
2. **Reversibility.** Preserved exactly. Residual cells are ordinary reversible replacements on the existing manifest; restore round-trips byte-for-byte in every test. No one-way primitive enters the core contract.
3. **Agentic-first.** Residuals ride the existing session, so multi-turn and staged callers get them without a new API, and strict-protection callers keep denying them the same way.
4. **Trust (auditable + deterministic).** Every residual is traceable to a rule preview and named parent IDs. The trace discriminator `ResidualPolicyTokenize` is private and projects to the existing `primary_pipeline / policy / tokenize` tuple, because that is what it truthfully is; it does not certify a whole entity. Safety-net provenance stays `safety_net`. No new scorer fields, no extra final trace rows, no fabricated `Candidate`.
5. **Adopter ergonomics.** Neutral in this PR. No public API, no new config, no new dependency.

## Immediate-caller flow

With `residual_coverage` off, every caller below is byte-identical to `main`:

- `gaze-cli` `pipeline/run.rs` and `commands/daemon.rs` — ordinary clean / pseudonymize
- `gaze-document` `bundle/mod.rs` — OCR text clean
- `gaze-document` `mcp/mod.rs`, `gaze-mcp-core` `dispatch.rs` — `protect_text_transaction`
- token-bridge `ingest.rs` — ordinary clean

`coverage_accounting_names_the_excluded_component_and_its_uncovered_bytes` pins that equivalence directly: with coverage disabled the manifest matches Stage A exactly.

## Falsification evidence

21 tests. The ones worth naming:

- **Pair and triple through the real public traced entry point.** `clean_text_with_safety_net_policy_detect_context_and_protection_trace` — the exact call the benchmark uses. Pair emits `0..15` + `15..21`, triple emits `11..15` + `15..16` + `16..29`, the whole token is byte-identical to Stage A, the adjacent quote is untouched, and restore is exact.
- **The unchanged scorer contract, asserted rather than printed.** The wire fixture now mirrors `gaze_bench_score.py:381-396` and `:663-749` clause for clause: exact key sets, the four allowed stage/decision/action tuples, canonical class, non-empty sorted duplicate-free source IDs, in-bounds char-boundary geometry, sorted disjoint trace spans, tokenize/manifest agreement by multiplicity, and no out-of-vocabulary source ID equal to protected raw text. Three mutants confirm it bites, including one that fails at the two-net fallback-deletion call site.
- **Static recognizability across the production registration hop.** The benchmark reaches this engine through `gaze_assembly`, which forwards rules as a generic `R: Rule + 'static`. If that hop ever stops preserving the preview, the planner admits nothing and the feature protects zero bytes in production while every other test stays green. A test now reproduces that hop and pins the coverage it must produce. Erasing the preview turns all 20 tests red; dropping `ColumnRule` from the downcast chain turns exactly 2 red.
- **Coverage accounting for excluded components.** One input carrying an admitted component and an excluded one. Observed: `covered=[0..15, 15..21, 30..45]`, excluded union `30..51`, uncovered `45..51`. It fails if coverage silently widens to the excluded component or silently narrows on the admitted one.
- **Fail-closed ledger.** Six corruption modes — popped parent, shifted bounds either way, forged ownership, out-of-range residual index, wrong action — all rejected by `Ledger::validate`.
- **Effects and ordering.** Allocation failure lands after the original effects and before the residual audit; audit failures preserve original order; no false rollback.
- **25 action pairs**, unknown before and after a match, wrapped built-ins, clone parity, and no eager custom rule calls.
- **Normalization.** Fullwidth, per-scalar NFC and removed joiners, with the selected raw-overlap guard unchanged.

## Operation and allocation bounds, honestly

Measured by in-test counters on the maximal-incidence fixture (129 originals):

- `preview_queries = 2` — the closed preview is memoized per class, so it is `O(distinct classes)`, not per cell.
- `active_parent_visits = 4160` and total cell parent incidence `4160` for 65 cells.

**`I`, the output parent incidence, is quadratic in the worst case** and the fixture demonstrates it rather than hiding it. There is no cap and no latency theorem here. The sweep is `O(n log n + I)`. Retained payload identity is asserted by pointer across four projections, so there is no per-cell candidate clone, no whole-clean-snapshot per suspect and no per-cell session-vocabulary scan. No timing claim is made in this PR; the paired benchmark is a separate, separately authorized run.

## Limitations, stated plainly

- **Off by default.** This PR protects no additional byte in production. The consumer migration and default flip follow.
- **Bytes outside `U` stay uncovered.** `U` is admitted original evidence only. Source bytes that no recognizer ever proposed are not protected by this work and are not claimed to be. The accounting test names such bytes explicitly.
- **Mixed and unknown components keep Stage A exactly**, including frozen Preserve spans. A single member whose effective class previews non-Tokenize excludes its whole component. The conservative contract requires both the original policy and its real standalone fallback policy, so eligibility is narrower than "all effective-Tokenize cases"; that is a disclosed limit, not a claim of completeness.
- **The selected raw-overlap denial is unchanged.** Two distinct normalized selections can map to one physical scalar (U+0344) and that input is still refused. This PR does not touch that guard, and that expected rejection must not be counted toward any zero-reject goal.
- **`I` is quadratic** as described above.

## Tracking

No open tracking item covers this work, so there is no `Resolves` line.

## Forward integration onto main

Rebased onto `main` at `77124be` (tree `0044fd56`, containing 587/588/589/590/591/592) with
`git rebase --onto origin/main e115f2c`, replaying 23 commits with **zero conflicts**.

A plain merge would have reported nine `add/add` conflicts, which are ancestry loss rather than
content disagreement: the branch's merge-base is `5260cf0` (#586) while 589 and 591 reached main as
squashes under new ids. Giving rebase the explicit upstream avoids that entirely.

Verified structurally rather than by reading the result:

- all 23 replayed commits are signed (`%G?` = `G` for every one) and carry their DCO trailer;
- the 19 files this work touches are **byte-identical** to the pre-rebase tree;
- the rebased head differs from `main` in **exactly those 19 files and nothing else**, so nothing was
  smuggled in or dropped by the replay.

Gates re-run on the rebased tree, since 591 touches residual batch handling in the same file and a
clean replay does not prove semantic compatibility:

| gate | exit | log sha256 |
| --- | --- | --- |
| `cargo fmt --all -- --check` | 0 | `e3b0c44298fc1c14…` (empty) |
| `cargo clippy --workspace --all-features --all-targets -- -D warnings` | 0 | `af0160ab54bbdaa3…` |
| `cargo test -p gaze-pii --all-features` | 0, 521 passed | `a5f1f6060d55f24f…` |
| `cargo test -p gaze-types -p gaze-token-bridge -p gaze-document --all-features` | 0, 183 passed | `2689a00aaa5005b2…` |
| `cargo test -p gaze-recognizers --test explicit_field_collision_control` | 0 | `6bf84ab1fc8044ac…` |

The production activation proof still holds against real main:
`field_present=true; whole=0..15 class=Name; residual=15..21 covering " right"; exact_restore=true`.

https://claude.ai/code/session_015mDhaTYfZyr3M9aYHdBML5


Naoray:
Superseded: merged into main via #597 squash c11e1880385e451cc52261aade39d13b6acd3bf8; branch head de75fa6f is byte-identical to the corresponding commits in #597.

PR #583, opened 2026-09-14
Prepare 0.15.0 with fresh release benchmark evidence

Naoray:
## Summary

Prepare 0.15.0 from merged main `3c341a3fabb0f98dc798fffd1e7a2cfe95dc0e5f`, including #584 final fallback admission and #585 configured-net proxy admission. Keep coordinated versions for all 15 published packages, bounded lock refreshes, current installation examples, and the complete v0.14.0 release-range changelog and migration guide. Runtime source matches merged main; this preparation adds release metadata, documentation and benchmark evidence.

Both source breaks remain explicit: fallible `PiiClass::custom` and `gaze_mcp_bridge::config::SessionCfg::max_sessions` (TOML default 1000). Migration guidance covers full rescanning, capacity/persistence, restore boundaries, agent/operator surfaces and subprocess diagnostics. It now also explains new configured-net denials and inference cost, missing/skipped-net coverage limits, unchanged primary Preserve/Redact and legacy clean defaults, and legacy mappings remaining live on denial. Direct failures discard staging before commit/send; snapshot admission is not whole-request rollback or serialization.

## Evidence and release status

**BLOCKED: full harness exit 3; release readiness FAILED (8 gates); regression FAILED (31 findings).** The fresh run measured signed clean D `7238894097054027850df56eb8aa5aa9497dbc2b` after both merged runtime fixes. The [v0.15.0 scorecard](docs/reference/benchmarks/scorecard-v0.15.0.json) retains that revision and `gaze.dirty=false`; SHA-256 `ac77709e0550cd4ebf92c6be1e3c6b1a3e7cccfcb747a221a1b42a7d706cf23c`. The [provisional row and charts](docs/reference/benchmarks/README.md) repeat failed-readiness and changed-population caveats beside both charts.

Production completed **2,820/2,910 (96.9072%)**, with **90 failed-closed documents**, all at `clean`: `safety_net_fallback_residual_suspect` **5**; `safety_net_invalid_output_other` **85**. The existing scorer recorded and reconciled their IDs, closed reasons/stages and exclusion digests. These are denials, not successful zero-leak documents. The reason bucket `safety_net_invalid_output_other` is retained as recorded; it is not a root-cause diagnosis.

[...]

PR #600, opened 2026-09-16
Prepare 0.15.0 release (r2)

Naoray:
## Summary

Rebuild the 0.15.0 release scaffold on current `main` (`3297b4da`). This replaces PR #583, whose measurement (commit D `7238894`) predates #586–#598 and whose probe lockfile predates #594. Nothing here changes runtime source: the diff is version pins, lockfiles, install examples, the curated CHANGELOG and migration guide, and the provisional-row renderer plus its tests.

Two things are deliberately **held** and are not in this branch — see *HOLD* below.

## What changed

- **Version pins.** All 15 crates in `cargo run -p xtask -- publish-plan` plus the `[workspace.dependencies]` pins move to `0.15.0`. Re-derived from the plan rather than copied from #583; the crate set matches.
- **Lockfiles.** `cargo update -w` refreshed `Cargo.lock` in 15 lines with **no dependency churn**. The `scripts/bench/validator_recall_probe` lockfile was regenerated **in place on top of #594**, not copied from #583: that branch predates #594, so re-applying its blob would have reverted the `windows-sys` entries and reintroduced the hard-coded `--locked` probe build failure that blocked the benchmark harness entirely. Result is 4 lines instead of 20.
- **Install examples.** Root and per-crate README version strings. Re-applied by hand against `main`, because `git diff main..agent/release-0.15.0` also shows five merged PRs' docs (#589, #590, #592, #597) as deletions; cherry-picking the blobs would have reverted them.
- **CHANGELOG 0.15.0 section**, curated over `v0.14.0..main` (51 merged PRs), Keep a Changelog buckets. Both source breaks retained and still true (`PiiClass::custom` fallible; `SessionCfg::max_sessions`).
- **UPGRADE.md** v0.15.0 migration section rewritten for the same range; the standalone "Pending security fix: prefix reuse disabled" preamble is folded into it.
- **Renderer.** #583's provisional-row changes to `render_benchmark_doc.py` and its 32 new tests. Every new branch is gated on `entry.get("provisional")`, and no row on `main` carries it, so the rendered document is byte-identical — proven by `--check` passing against the unchanged history.

### CHANGELOG entries worth reading

- **Security, disclosed as a shipped v0.14.0 defect (#593).** Both `gaze-proxy` residual checks decided on surviving manifest entries alone. A safety-net fallback deletion emits no manifest entry, so a deleted net-only span returned `Ok` with an empty span list, which both call sites read as "no PII found" while still holding the raw bytes. Reproduced before the fix on three surfaces: an unsurfaced request field forwarded verbatim (200), a buffered JSON response, and a per-`text_delta` SSE response. Axis 1 in its literal form.
- **Residual coverage on by default (#595, #597)** is called out as a token-stream change for *every* adopter: a manifest span count is now a count of replacements, not of distinct recognized values. The token-bridge consequence is stated plainly — a residual fragment is protected but unsearchable.
- **Postal `ca`/`gb`/`ie` (#598)** with the documented residual: they are `locale_basis = "format"`, so they run at every locale including `--locale=global` and **cannot be suppressed with a locale chain**; an adopter who must not tokenize them has to disable the recognizer.

## HOLD — both now cleared

~~1. Benchmark row and scorecard~~ — **done.** The benchmark lane landed the **non-provisional** `v0.15.0` row and `scorecard-v0.15.0.json` on this branch at `48ac1bb`, measured on runtime commit `9a3a788`. Gate 9 is satisfied and `render_benchmark_doc.py --check` is in sync against the history that now contains the row. Nothing is held any more.
~~2. #599 (hybrid terminal admission)~~ — **done.** #599 merged at `9a3a788` and both placeholders are filled; see *Merging #599* below. The benchmark row above is now the only thing held.

The release date in the CHANGELOG heading is `2026-09-16` and must match the day the tag is actually cut.

## Pre-flight gates

| # | Gate | Result |
| --- | --- | --- |
| 3 | Version pins across workspace + per-crate manifests | **PASS** — 15/15 publish-plan crates at `0.15.0`, exit 0 |
| 4 | `ci-feature-matrix` | **Not run locally, by design** — CI's `xtask gates` job. It was **14/14 green on the pre-merge head `c7da44a`**, and is re-running on the merge |
| 5 | `cargo run -p xtask -- readme-version-check` | **PASS**, exit 0 — 15 published crates plus both accept/reject fixture tests |
| 6 | Absolute-path grep over tracked files | **PASS** — zero matches, exit 0 |
| 7 | Dogfood (below) | **PASS with a known non-PII class** |
| 8 | Benchmark claims cite script + hardware | **PASS** — the postal measurement now cites the harness invocation, the seed, the corpus, the per-release hardware field, and the test that pins the accepted false-positive shape |
| 9 | Fresh benchmark paired with the release | **PASS** — non-provisional `v0.15.0` row + committed scorecard, measured on `9a3a788` |

Renderer: 65 tests pass; `render_benchmark_doc.py --check` reports the document in sync.

## Dogfood proof (gate 7)

Both release texts were piped through the `gaze clean` built from this branch.

| text | detections | classes |
| --- | ---: | --- |
| CHANGELOG `0.15.0` section | 12 | `custom:ip_address` ×12 |
| UPGRADE `v0.14.x → v0.15.0` section | 18 | `custom:ip_address` ×18 |

Identical counts under `--rulepack-bundled core` and `core-extended`.

**Every one of the 30 detections is `ip.v6` matching a Rust scope separator** — `Pipeline::builder`, `PiiClass::custom`, `EmittedTokenSpan::new` — which is the already-tracked todo #2402, documented on the `ip.v6` row in `redaction-classes.md` and reproduced here on our own output, exactly as #598's dogfood reported. **Zero PII, and no detection of any other class.** The accurate API paths are kept because distorting real symbol names would cost adopters more than the noise does.

Restore round-trip on the tokenized CHANGELOG section is **exact**: 18,788 bytes in, 18,788 bytes out, byte-identical.

Nothing shaped like a postal code, email, name or phone number fired on either text.

## Merging #599

#599 appended to `[Unreleased]`, which this branch had moved wholesale into `[0.15.0]`, so the merge conflicted structurally rather than textually — stripping markers would have silently dropped its 43 lines. The resolution takes this branch's heading structure and re-homes #599's two bullets into `[0.15.0] ### Changed`, verbatim apart from two `(#599)` citations.

Verified by reconstruction rather than by reading:

- **+43 / −15 against this branch's own parent** — exactly #599's block in, exactly the placeholder comment out, nothing else.
- **40 of #599's 42 non-blank CHANGELOG lines are byte-identical**; the two exceptions are the lines that gained the citation.
- **Zero conflict markers of all four zdiff3 types** across the tree.
- A sweep of every line the resolution dropped relative to `origin/main` found **one fact the curated rewrite had summarized away**: the embedded `core` rulepack version `0.5.3`. Restored as its own entry. Everything else removed is a deliberate rewrite of the same fact, checked individually.

`UPGRADE.md` gains a matching subsection that leads with the widening — nothing that completed under v0.14 starts denying — then names the two costs: one extra model pass on a fallback document that reports at the terminal scan, and a fresh post-round finding shipping raw with an honest report. That last one is **2 bytes in one document, overlapping 0 gold**, out of the 42 bytes across 16 spans that previously denied, stated as a measurement on that corpus, seed and model bundle rather than as a bound — and a shipped byte that overlaps no gold is not proof it is not PII, only that the benchmark does not count it.

1. **Verdict:** `skip`.
2. **Opportunity:** `none`.
3. **Why:** the change carries no runtime state, no branching and no duplicated rules. Version pins are derived from the publish plan, not enumerated; the rendered benchmark document stays a projection of `release-history.json` through the single existing renderer; the provisional-row logic added here reuses one `_display_version` helper rather than repeating the suffix at each of its four call sites, which is itself the small cleanup the change earns.
4. **Scope:** reviewed the changed manifests and their immediate consumers (both lockfiles, the publish plan, `readme-version-check`), the renderer and its 65 contracts, and the two release documents. No abstraction is worth adding for a release scaffold that is regenerated once per minor.
5. **Validation:** gates 3, 5, 6, 7, 8 above with their exits; renderer unit tests and `--check`; exact restore round-trip; `cargo update -w` diff inspected line by line for churn. Gate 4 runs in CI on this PR; gate 9 is the benchmark lane's.

## Five axes

- **Reliability:** the headline entry is a disclosed shipped leak and its fail-closed fix; residual coverage on by default widens protection for every adopter; the postal rules close the largest single leak bucket.
- **Reversibility:** unaffected, and demonstrated — the release text round-trips byte-exact through `clean` and `restore`. #588 and #591 are described as widening reversible coverage before any one-way deletion.
- **Agentic-first:** the migration guide leads with proxy and MCP surfaces, the new denials an agent host will see, and the extra inference cost they carry.
- **Trust:** every measured number cites its script, seed, corpus and hardware field; the dogfood result is reported as it actually is rather than rounded to "zero"; the held benchmark row is named rather than quietly provisional.
- **Ergonomics:** two source breaks, the token-stream change, and the unsuppressable postal rules are each stated with the action the adopter has to take.

## The benchmark result, including the part that went the wrong way

Measured across all 2,910 production documents on a scored-document digest byte-identical to v0.14.0 (14,719 gold entities / 130,282 gold bytes), so this is like-for-like rather than an estimate:

| metric | v0.14.0 | v0.15.0 | |
| --- | ---: | ---: | --- |
| surviving PII bytes | 25,179 | **22,491** | −10.68% |
| leak rate | 19.327% | **17.263%** | −2.06 pp |
| exact restoration | 78.42% | **96.53%** | |
| failed closed | 11 | **0** | |
| clean p95 | 195.86 ms | **251.36 ms** | **+28.3%** |

**The p95 regression is real and is disclosed rather than traded away.** It is the cost of full input rescanning plus the extra configured-net and terminal model passes, and it appears in the CHANGELOG known limitations, the UPGRADE.md TL;DR, and its own migration subsection telling a p95-sensitive adopter to measure before rolling out — the correctness changes that cause it are not individually opt-out. **22,491 labeled PII bytes still survive**, so both documents say measured improvement, not completeness.

## Not done here

No tag, no publish, no merge. Gate 4 is CI's; gate 9 is the benchmark lane's; tag authority is the user's.

---

https://claude.ai/code/session_01JPfHnECE2o22n1sJwUPzte


Naoray:
HOLD (maintainer decision 2026-09-16): no 0.15.0 release until the detection design is credible. Shipped-defect found in the Kiji safety-net decoder (WordPiece-level spans, mid-word tokens) and net label confusion; fix + re-benchmark + safety-net replacement leaderboard come first. Keep this PR as the scaffold; do not merge.

Naoray:
Closed while the release HOLD stands (user 2026-09-17: no long-lived open PRs). The 0.15.0 release PR will be regenerated from main when the user lifts the hold; branch agent/release-0.15.0-r2 is kept.

PR #601, opened 2026-09-16
Name DetectError at the gaze crate root so out-of-crate recognizers compile

Naoray:
## What

`Recognizer::detect` returns `Result<Vec<Candidate>, DetectError>`. Every other type an adopter needs to write that signature was already re-exported at the `gaze` crate root; `DetectError` was not, so an out-of-crate impl had to reach for `gaze::registry::DetectError` or take a direct `gaze-types` dependency. The latter is impossible for `gaze-token-bridge`, whose `Cargo.toml` is a frozen contract, so `crates/gaze-token-bridge/tests/residual_fragment_ingest.rs` works around it with the module path today.

- `crates/gaze/src/lib.rs` — add `DetectError` to the crate-root `pub use registry::{...}` list. Purely additive; no behavior change.
- `crates/gaze/src/registry.rs` — module rustdoc with a complete out-of-crate `Recognizer` implementation, written entirely in `gaze::` paths.

The example is a **doc-test**, which compiles as a separate crate. That is the only kind of test that fails when the public surface is unusable from outside: a unit test inside `gaze` would compile even with the re-export missing. It also exercises `Candidate::new` — `#[non_exhaustive]` makes that the sole construction path for adopters — and shows the fail-closed error return via `DetectError::backend`.

Closes the ergonomics papercut filed as todo 3632.

## Axes touched

- **Axis 5 (adopter ergonomics)** — improved. One missing name forced adopters onto an internal-looking module path.
- **Axis 4 (trust)** — the example teaches the fail-closed contract: a backend that cannot scan returns `DetectError`, never an empty candidate list that would read as a clean document.
- No axis weakened. No runtime behavior changes.

## Immediate callers

`pub use` additions have no callers to update. The existing `gaze::registry::DetectError` path still resolves, so the `gaze-token-bridge` test that works around this keeps compiling unchanged; it is left alone rather than churned in a hygiene PR.

1. **Verdict:** `implement` (already implemented here).
2. **Opportunity:** `none` beyond the re-export itself. The crate root is already the single flat export surface; this restores its completeness rather than adding structure.
3. **Why:** removes a concrete adopter dead end (`error[E0425]: cannot find type DetectError in crate gaze`) without adding indirection.
4. **Scope:** two files in `crates/gaze`, additive only. Nothing waits.
5. **Validation:** commands below, all exit 0.

## Commands

| Command | Exit |
|---|---|
| `cargo test -p gaze-pii --all-features` | 0 (includes the new doc-test: `crates/gaze/src/registry.rs - registry (line 8) ... ok`) |

Gate log sha256: `ccbe5f6e606e103520fbbcc1d29acf9d67aa0ea836452cef5c24fcfbfab89528`


Naoray:
pr-drain (brief 9964): `xtask no-tenant-knowledge` failed because the doc example recognized an order identifier (`order_id` is denylisted). The gate's intent is that core source carries no tenant business vocabulary, so the example now recognizes colon-separated **MAC addresses** (`00:1a:2b:3c:4d:5e`): a hardware identifier defined by structure alone, not a renamed business-entity id. Denylist and gate untouched. It still shows every point of the original: `gaze::`-only paths, `Candidate::new`, and the fail-closed `DetectError::backend` return.
Merged current main (signed, signed-off); fix commit 1b6c7e51.
Checks: `no-tenant-knowledge` passed; doc-test ok; fmt ok. Mutation: dropping `DetectError` from the crate-root `pub use` → doc-test FAILED with `E0432 unresolved import gaze::DetectError`.

PR #602, opened 2026-09-16
Run the origin-agreement guard on every record, not once per segment

Naoray:
## What

`Ledger::validate` checks that the public `EmittedTokenSpan.origin` agrees with the internal `Origin`. It read no `segment`, yet sat inside `for segment in &self.segments`. Two consequences, both fixed here:

1. **Redundant, and quadratic where it matters.** `existing_owned` pushes one segment per carried owned token, and both restore-boundary protection paths call it in a loop before `spans.validate()`. There segments and records both grow with the number of protected tokens, so the guard was O(K²) in tokens on the protect path. Single-shot `clean` has one segment, so this was never a benchmark concern.
2. **Skipped entirely when `segments` was empty.** A fail-closed check whose execution is gated on an unrelated collection being non-empty is fragile by construction. Reachability was traced rather than assumed: no production site reaches it today, so this was **latent**, not a live leak.

**The mechanical hoist would have been a regression.** Records built through `From<Vec<EmittedTokenSpan>>` all carry `Origin::Unknown`, and the predicate conflates "unknown" with "not a fragment", so a legitimately emitted fragment re-imported that way would have newly failed validation. The guard therefore skips `Origin::Unknown` and fires on every origin that *is* known — strictly stronger than the nested version, and safe to hoist.

Implements todo 3629 per its Definition of done.

## Tests, and the probe that proves they are not decorative

Both new tests live in `occurrence.rs`'s own test module, where `Ledger`'s private fields are visible.

| Mutation applied to the committed tree | Result |
|---|---|
| Revert the hoist (guard back inside the segment loop) | `the_origin_agreement_guard_runs_on_a_ledger_with_records_and_no_segments` **FAILED**, re-import test passed |
| Hoist without the `Origin::Unknown` skip | `a_fragment_reimported_through_the_span_constructor_still_validates` **FAILED**, no-segments test passed |

Each mutation reddens exactly one test, so neither test is redundant and neither survives the other's mutation.

The no-segments test asserts on the **error message**, not on `is_err()`. That is load-bearing: probe 1 showed the pre-hoist code still returns *an* error for that ledger — `manifest-integrity: invalid reconstructed ownership`, from a later check that looks the segment up. An `is_err()`-only assertion would have passed against the broken code and pinned nothing.

`a_record_whose_two_origins_disagree_fails_closed` (the existing test, built via `Ledger::new(segment)` with S == 1) stays green.

## Axes touched

- **Axis 1 (reliability) and axis 4 (trust)** — strengthened. The fail-closed check now runs on every record regardless of segment count, and covers `Selection`, `SafetyNet` and `ExistingOwnedUnknown` origins rather than being reachable only through a segment-bearing ledger.
- **Axis 2 (reversibility)** — protected. Skipping `Origin::Unknown` is what keeps a re-imported fragment restorable instead of rejected.
- Performance improves as a side effect (one O(R) pass instead of O(S×R)); no axis is weakened for it.

## Immediate callers

`Ledger::validate` is called from `crates/gaze/src/pipeline/protection.rs:136` and `:259` (both restore-boundary protection paths), from `merge` before splicing, and from the pipeline's manifest checks. All were reviewed: each pushes its segments before validating, so none relied on the segment-gated behavior and none needed a change. No caller observes the origin guard's position, only its verdict.

1. **Verdict:** `implement` (done here).
2. **Opportunity:** `none`. The organizing idea already exists — one `Origin` enum as the single source of truth for the public discriminator. This restores the invariant's enforcement to match that shape instead of adding structure.
3. **Why:** removes a duplicated O(S×R) pass, removes an invalid intermediate state (a fail-closed check that silently does not run), and makes the "unknown ≠ not-a-fragment" distinction explicit in code rather than accidental in control flow.
4. **Scope:** one function plus two tests in `crates/gaze/src/pipeline/occurrence.rs`. Nothing else should wait on it.
5. **Validation:** commands below, all exit 0, plus the two mutation probes above.

## Commands

| Command | Exit |
|---|---|
| `cargo test -p gaze-pii --all-features` | 0, zero `test result: FAILED` lines |

Gate log sha256: `c2ed7b7892f11a4777d3d08f981e71c8c3b2505340de2ddcdd916c3da1cc2e27`


Naoray:
Independent verification (pr-drain, brief 9964):
- `Origin::Unknown` is only constructed via `Occurrence::unknown`, reached from `From<Vec<EmittedTokenSpan>>` and the test-only `push`. Gating that `From` impl behind `#[cfg(test)]` still compiles `gaze-pii --lib --all-features`, so no production ledger holds an `Unknown` record: the skip removes no production rejection.
- Mutation 1 (delete the `Unknown` skip): only `a_fragment_reimported_through_the_span_constructor_still_validates` FAILED.
- Mutation 2 (guard runs only when `segments` is non-empty): only `the_origin_agreement_guard_runs_on_a_ledger_with_records_and_no_segments` FAILED.
- Merged current main in (signed, signed-off). Verdict: CLEAN.

PR #603, opened 2026-09-16
Remove the unverified core.toml line-citation column from the recognizer matrix

Naoray:
## What

The recognizer coverage matrix in `docs/reference/redaction-classes.md` carried a thirteenth column citing `crates/gaze-recognizers/embedded/core.toml:<lo>-<hi>` per recognizer. `crates/xtask/tests/redaction_classes_doc.rs` builds `RecognizerRow` from `cells[0..12]` and **never reads `cells[12]`**, so the column was documentation that no gate verified. At PR 598's head, 35 of 37 rows cited stale ranges, one off by more than 150 lines (`security_token.anchored` cited `762-847`, actually `985-1043`).

Two ways out. Verifying it means encoding a block convention — where a recognizer's TOML starts, and where it ends minus the trailing comment run that introduces the next one — that nothing else in the codebase needs, purely to protect a field no code reads. Regenerating without verifying, as an earlier commit this cycle did, only resets a clock that drifts again on the next `core.toml` edit.

So the column goes, per the decision on todo 3636: keep gates that verify active code, drop documentation cosmetics that nothing depends on.

- `docs/reference/redaction-classes.md` — column removed from the header, separator and all 44 rows; prose now names the two TOML files and the `id = "..."` line to search for, which is stable in a way a line range is not.
- `crates/xtask/tests/redaction_classes_doc.rs` — expected column count 13 → 12.
- `CHANGELOG.md` — the earlier "citations regenerated, assertion tracked separately" entry is replaced by the removal, so the released history does not claim a column that no longer exists.

**The gate itself stays and is unchanged in what it verifies**: documented recognizer rows against the loaded rulepack — id, matcher, class, locales, validator, normalizer, safety tier, `safe_default`, base score and priority.

## Proof the gate still bites

Not a presence check. With the column removed, typoing one recognizer id in the table (`email.global` → `email.globel`) reddens the gate:

```
test redaction_classes_reference_matches_loaded_code ... FAILED
            id: "email.global",
            id: "email.globel",
test result: FAILED. 2 passed; 1 failed
```

The column count is still asserted too, so nobody can silently add or drop a column.

## Scope note — four sibling unchecked columns, reported not deleted

Every gated table in this file has a trailing column the gate parses and discards: `pii-classes` (`cells[3]`), `validators` and `normalizers` (`cells[3]`, `cells[4]`), `collisions` (`cells[5]`), `default-activation` (`cells[4]`). They carry the same `path:<lo>-<hi>` shape and drift on the same mechanism.

They are **not** touched here. The 3636 decision named this one column, some of those columns point at several files and carry navigational value the recognizer column did not, and the right treatment differs per table. Filed as todo 3647 for one decision across all five rather than a mechanical sweep.

## Axes touched

- **Axis 4 (trust)** — improved. A column that looked verified and was not is the dormant-gate pattern; removing it means everything the table shows is now gate-checked.
- **Axis 5 (adopter ergonomics)** — net neutral. A reader loses a line range that was wrong 95% of the time and gains a stable way to find the same definition.
- No axis weakened. No code behavior changes.

## Immediate callers

`redaction_classes_doc.rs` is the only reader of this table; `DOC_PATH` is its single reference. `CLAUDE.md` links two other sections of the document (`#closed-validator-and-normalizer-sets`, `#shipped-default-activation`), both untouched, and no anchor changes.

1. **Verdict:** `implement` (done here).
2. **Opportunity:** `none` — the correct move was removing a field, not introducing a structure. A line-range verifier would have been exactly the speculative abstraction the rule warns against: new machinery, no invariant encoded, protecting data nothing reads.
3. **Why:** removes a source of silent documentation drift and one instance of the dormant-gate pattern, with no indirection added.
4. **Scope:** one doc, one test constant, one CHANGELOG entry. The four sibling columns should wait for their own decision (todo 3647).
5. **Validation:** commands below, all exit 0, plus the mutation probe above.

## Commands

| Command | Exit |
|---|---|
| `cargo test -p xtask --all-features` | 0, zero `test result: FAILED` lines |

Gate log sha256: `36d2b0546f50ca3f2dfc5bdeb01c8e3f6d4a5bee661e0d5544b595bbf0f49154`


Naoray:
pr-drain (brief 9964): merged current main (2f6dc923, signed + signed-off). Conflict in `docs/reference/redaction-classes.md` resolved by taking main's table (PR 607 moved `security_token.anchored` / `password.field` / `username.field` to `secrets`) and dropping only its last column; all 41 dropped cells were `crates/gaze-recognizers/embedded/...` citations. Prose + CHANGELOG no longer name the nonexistent `core-extended.toml`: `core-extended` is an alias that loads `core.toml` (`crates/gaze-recognizers/src/lib.rs:54`), and the other file is `secrets.toml`.
Gate: `cargo test -p xtask --test redaction_classes_doc` 3/3 ok. Mutations: id typo in a `secrets` row → `redaction_classes_reference_matches_loaded_code` FAILED; re-adding a 13th header cell → same test FAILED.

PR #613, opened 2026-09-17
Add an anchored four-digit postal rule for Austrian and Swiss documents

Naoray:
Adds `postal.at_ch`, an anchored four-digit postal-code recognizer for Austrian (`de-AT`) and Swiss (`de-CH`) documents. Resolves todo 3633.

## Why

Four-digit codes in de-AT / de-CH documents had no recognizer: 366 gold ZIP entities (1,470 bytes) in the EN/DE holdout leaked in full. ZIP is the second-largest leak pool. A bare four-digit string carries no structural signal (unanchored `\d{4}` is 19% precise on the holdout and fires across 62.5% of the negative corpus), so this rule matches only:

* directly **after a postal cue** (`PLZ`, `Postleitzahl`, `Postcode`, `ZIP`, `Zip code`), or
* directly **before a city-shaped token** (uppercase start plus two letters, or `St.`/`St` before a capitalised name), across an optional comma and 1 to 3 spaces, NBSP or NARROW NBSP.

An `A-` / `CH-` / `FL-` prefix is part of the token. `locale_basis = "document"`, `locales = ["de-AT", "de-CH"]`, `safety_tier = "locale_gated"`, cue baked into the regex with `capture_groups = [1, 2]` (`anchored_match` cannot match digits).

## Measured (base `9f1cd524`, head `0c883b7c`, full 2,910-document population, both contracts, `dirty = false`)

| contract | arm | ZIP fully covered | leaked bytes | FP bytes |
|---|---|---|---|---|
| v1 | `rule-floor-extended` | 548 → **862** | 94,302 → 93,032 (**−1,270**) | 5,433 → 5,545 (+112) |
| v1 | `pass2-ner` | 548 → **862** | 27,321 → 26,051 (**−1,270**) | 28,045 → 28,157 (+112) |
| v2 | `rule-floor-extended` | 548 → **862** | 87,647 → 86,377 (**−1,270**) | 5,433 → 5,545 (+112) |
| v2 | `pass2-ner` | 548 → **862** | 20,727 → 19,457 (**−1,270**) | 28,045 → 28,157 (+112) |

ZIP byte recall on the rule floor goes from 59.7% to 83.0% (+1,262 ZIP bytes). No document failed closed. Timings are provisional (shared machine) and not claimed.

**Known cost, disclosed per the user decision of 2026-09-16:** **28 tokens** on the 567 de-AT/de-CH holdout documents overlap no gold span (+112 FP bytes). 16 follow a postal cue and are most likely unannotated codes. 12 follow a capitalised word: German capitalises every noun, so `1500 Euro` looks like `1500 Musterstadt`, and the `regex` crate has no negative lookahead for a stop-list. All restore losslessly. 3 more cover the year of a date of birth. **0 / 1,024** negative documents, both as committed and with every document forced to de-AT or de-CH.

## Proof

* **TDD:** 21 fixtures in `postal_at_ch.rs`, written first: 14 of the first 19 failed on base, all passed after the rule landed; 2 more were added for the enumeration findings. Synthetic town names only.
* **Differential enumeration, both directions:** 302,400 inputs (9 code shapes × 25 left contexts × 12 separators × 16 followers × 7 locale chains) through the real `clean_for_bench` binary, base vs head. **0 inputs where base protected a byte the head leaks. 0 changes outside de-AT/de-CH chains. Every restore exact.** The enumeration found two precision classes the corpus could not see, both fixed at zero gold cost: `1500 St.` ("1500 pieces") and `#4711 Fehler` (issue reference).
* **Mutation roster: 18/18 killed**, including: drop the anchor, widen to five digits, add `de-DE` to the locale gate, drop the cue branch, drop the `#` guard, drop the prefix, drop NBSP, exclude years, drop the doc activation row.
* **Locale chain pinned:** `[de-AT]`, `[de-DE]`, `[de-AT, de-DE]`, `[de-DE, de-AT]` and the fall-through case.

## Adopter-visible behaviour

* **First-locale-wins.** With `[de-AT, de-DE]`, a document holding an Austrian code stops at de-AT and `postal.de` never runs, so a German five-digit code in the same document stays untokenized. That is the same trade `postal.de` / `postal.us` already make; it is pinned by test and documented. Use one country locale per document.
* **`core-extended` no-policy compatibility chain** (`global, en-US, de-DE, de-AT, de-CH`) now reaches this rule for documents where `postal.us` / `postal.de` found nothing. Forced onto the 1,319 other holdout documents it produced 54 gold ZIP, 142 other-gold and 10 no-gold tokens. `--locale=global` or a narrower policy avoids it.
* `en-AU` / `en-NZ` stay uncovered: postcode follows the locality there, so this anchor does not transfer.

1. **Verdict:** `recommend` (nothing implemented beyond this rule).
2. **Opportunity:** class-level cooperation for `custom:postal_code`, via a `postal` collision family or one class-wide cooperation declaration, instead of six hand-maintained `cooperates_with` lists.
3. **Why:** every new postal rule edits every sibling's list (n² data), and the loader mutation probe in `core_extended.rs` keys on those literal lists, so each addition also rewrites the probe. A class-level declaration would state the invariant once.
4. **Scope:** `core.toml` plus the loader's same-class check. It changes a rulepack contract, so it should wait for its own PR, not this one.
5. **Validation:** needed before adopting: the loader's `SameClassWithoutCooperation` fail-closed test must still fire for an undeclared same-class pair.

Also in this PR: the redaction-classes `core.toml:START-END` citation column was regenerated. Three rows (`ssn.de_cue`, `tax_number.cue_anchored`, `driver_license.cue_anchored`) were already stale on main after the secrets-pack move.

## Gates

Run on the rust 1.96.0 CI pin: `cargo fmt --all -- --check`, `cargo clippy --workspace --all-features --all-targets -- -D warnings`, `cargo test --workspace --all-features` (203 test binaries ok), and the xtask gates `bundle-tokenization-drift --verify-ack` (no snapshot diff: its corpus runs under `global`, where this rule cannot fire), `fixture-citation-lint`, `recognizer-composition-validator`, `family-policy-table-coherence`, `locale-cue-bundle-coherence`, `readme-version-check`, `no-tenant-knowledge`, `class-map-override-safety`, `symmetric-potemkin`. All green. The workspace run caught two clippy `from_ref` lints in the new test file; fixed, then clippy, fmt and the new test file re-run green.

Dogfood: this body and the CHANGELOG entry run through `gaze clean --rulepack-bundled core-extended` from this branch contain no PII. The only postal tokens are this PR's own examples (`1500 Euro`, `Musterweg 12 4020 Musterstadt`), a live reproduction of the disclosed compatibility-chain cost on an English document.

https://claude.ai/code/session_01NqcNxjGAtWdWe56pUKusfz


Naoray:
# Independent review — PR 613 (`postal.at_ch`, todo 3633)

Reviewer: independent Opus 5 delegate.
Subject: head `a18c2ea1`, base `9f1cd524` (= origin/main at review time), draft.

## VERDICT: BLOCKED

One recall-direction regression versus base, reproduced on real `clean_for_bench` binaries.
The brief classifies this as a BLOCKER, not a documented quirk. The fix changes detection
semantics (registry loop or class contract), so it is not a reviewer fix-now.

## B1 — BLOCKER (axis 1). A four-digit match, true OR false, switches off `postal.de` / `postal.us` for the whole document

`RecognizerRegistry::detect_candidate_pool` (`crates/gaze/src/registry.rs:583-603`) runs document-basis
recognizers per class, per chain locale, and `break`s at the first locale that yields any candidate.
On base, `custom:postal_code` has no de-AT/de-CH rule, so `[de-AT, de-DE]` always falls through to
`postal.de`. On head, any `postal.at_ch` hit at de-AT ends the loop and the German/US five-digit ZIP
that base tokenized now ships raw.

Reviewer enumeration (3,916 inputs, 11 chains, base vs head, same comparator
logic as the author's):

* **270 inputs where base protected bytes the head leaks.** 0 of them are single-code inputs; all are
  mixed documents. The author's 302,400-input grid placed exactly one code per input, so it could
  not observe this class.
* By chain: `de-AT,de-DE` 90, `de-CH,de-DE` 90, `de-AT,en-US` 90.
* By trigger (45 each): `4020 Musterstadt`, `CH-8001 Musterstadt`, `PLZ 1010`, and the **disclosed
  false-positive class** `1500 Euro`, `3000 Mitarbeiter`, `2024 Wien`.
* Lost spans: `10115` ×108 (incl. `D-10115`), `80331` ×54 (`PLZ 80331`), `62704` ×54, `90210` ×54.

Example (chain `[de-AT, de-DE]`): `Kosten laut Anlage. 1500 Euro. Lieferadresse: 10115 Musterberg.`
base tokenizes `10115`; head tokenizes `1500` (a false positive) and leaks `10115`.

Why this is not acceptable as a documented trade:

1. `[locale].active` is documented as the **locale fallback chain** (`docs/reference/policy.md:50`).
   `[de-AT, de-DE]` is the natural Austrian configuration, not a mixed-country misuse.
2. The PR's own docs (core.toml note, redaction-classes.md:358, CLAUDE.md primer, PR body) describe
   the shadowing as happening "in a document holding both shapes". That understates it: a price or
   headcount (`1500 Euro`) is enough. The advice "use one country locale per document" does not
   protect a single-country Austrian invoice that carries a German delivery address.
3. `locale_chain_first_locale_wins_against_postal_de_is_pinned` asserts the leak
   (`assert!(at_de.contains("10115"))`), so the test pins the regression rather than guarding against it.

Unaffected: the no-policy `core-extended` chain (`global, en-US, de-DE, de-AT, de-CH`) and the benchmark
(one locale per document), because de-DE / en-US precede de-AT there. That is why no scorecard moved
the wrong way.

Fix options (orchestrator / user call; each changes detection semantics):

* **A (recommended): registry span-aware fall-through.** Keep evaluating later chain locales for a
  class and admit their candidates only where they do not overlap an earlier locale's candidates.
  Removes this regression and the pre-existing `postal.de`/`postal.us` shadowing in one rule. Needs its
  own PR with base-vs-head enumeration across every class that has document-basis rules in more than
  one locale, plus a scorecard run.
* **B: move the four-digit rule out of the shared loop** (for example its own class with a shared token
  family). Narrow, but changes the class contract and audit rows.
* **C: ship P4 only after A lands**, rebased, with a mixed-document fixture that asserts BOTH codes are
  tokenized under `[de-AT, de-DE]`, `[de-CH, de-DE]`, `[de-AT, en-US]`, and the same with `1500 Euro` as
  the trigger.

## Confirmed TRUE (reproduced)

| Claim | Result |
|---|---|
| Single-code inputs: no base-caught span lost | TRUE on my grid: 0 regressions outside mixed documents |
| Restore exact | TRUE: 0 non-exact restores, both binaries, 3,916 inputs |
| de-DE / global / en-US gain no four-digit tokens | TRUE on single-locale chains |
| Scorecards (committed JSON, `dirty=false`, base `9f1cd524`, cand `0c883b7c`) | TRUE: v2 rule floor 87,647 → 86,377, pass2 20,727 → 19,457, v1 same Δ −1,270; ZIP covered bytes 3,237 → 4,499 (+1,262) |
| Per-label movement | ZIP +1,262 B; DATEOFBIRTH +8 B (the 3 disclosed DOB-year overlaps); **BUILDINGNUM +4 B, 54 → 58 fully covered, not mentioned in the PR body** (minor: disclose) |
| Author finding 4 (M4 fires nowhere) | Explained, not a P4 dependency: `recognizer_activates` (`gaze-assembly/src/detector_wiring.rs:297-309`) admits a `locale_gated` + `format` recognizer ONLY when `auto_activate_locale_gated = true`; the fixtures set it false. Under auto-activation it would fire everywhere. No bundled rule uses the combination; an adopter pack that does is silently never registered under an explicit-locale policy. Tracked separately. |

## Out-of-corpus shape observations (non-blocking, for the fix round)

Tokenized at de-AT (new precision classes beyond the disclosed noun class, all capitalised followers):
`um 1430 Uhr`, `Termin 1430 Uhr`, `Rechnung 1500 Offen`, `Version 2024 Release`, `Seite 1500 Kapitel`,
`1500 EUR`, `Betrag 1500 CHF`, `ab 1.1500 Stück`, `Bestellnummer 0815-0815 Artikel`.
Not tokenized (recall gaps, not regressions): tab or 4+ spaces before the city, `PLZ.: 1010`,
`PLZ -1010`, lowercase city, markdown cell `|1010|Musterstadt|`, city before code.
Correctly refused: IBAN groups, phone groups.

## Not done (deferred to the re-review of the fixed head)

Holdout FP list reproduction (item 4), mutation re-run (item 7), local workspace gates. The fix for B1
changes the head and likely the fixtures, so results on `a18c2ea1` would be discarded. CI on
`a18c2ea1` was still pending (test, xtask gates, MSRV, windows) at review time.

1. **Verdict:** `recommend`.
2. **Opportunity:** span-aware per-class locale fall-through in `detect_candidate_pool` (option A), and
   the author's class-level cooperation declaration for `custom:postal_code`.
3. **Why:** the `break` encodes "first locale wins for the whole document", which is only correct when
   rules of one class in different locales can never both be right in one document. Postal codes break
   that invariant. Span-aware fall-through states the real invariant (earlier locale wins per span).
   Class-level cooperation removes six n² `cooperates_with` lists.
4. **Scope:** A = `registry.rs` loop + tests + scorecard, its own PR, blocks P4. Cooperation = `core.toml`
   + loader same-class check, own PR, can wait.
5. **Validation:** needed: base-vs-head mixed-document enumeration across all multi-locale document-basis
   classes; `SameClassWithoutCooperation` fail-closed test must still fire.

## Five axes

1. Reliability: **regresses** for fallback chains (B1). 2. Reversibility: unchanged, restore exact.
3. Agentic-first: neutral. 4. Trust: the pin test asserts a leak; docs understate the trigger.
5. Ergonomics: disclosed FP cost is honest; the shadowing advice is misleading for the fallback-chain case.

PR #615, opened 2026-09-17
Split benchmark leaked bytes by whether the gold passes its own validator

Naoray:
## What

Gold identifiers that fail their own checksum stay scored gold (user decision on todo 3648, option b). This PR makes that explainable instead of hidden in the headline:

- **Scorer** (`gaze_bench_score.py`): every validator-backed label in `validator_recall_by_label` gains `production_recall_by_gold_validity`, the arm's own coverage split into gold that passes its validator and gold that fails it. It sits next to the existing validator-backed and shape-only recall and the gold pass/fail counts.
- **Runner summary**: two new columns, surviving bytes on valid gold and on invalid gold.
- **Benchmark doc renderer**: from the next measured release on, the current-release section adds a table for the shipped default arm (gold, gold failing its validator, validator-backed recall, shape recall, surviving bytes split). Rows recorded before the split carry no such field and render byte-identically (`--check` in sync).
- No corpus edit, no validator change, headline leaked bytes are never recomputed from the split.

## Evidence

**Scorecard identity (contract v2, full profile, seed 20260710, same machine):** base `9f1cd524` vs head `ab870b79`. A recursive JSON diff, ignoring only timing blocks, finds **54 added fields, all `production_recall_by_gold_validity`**, and one changed value, `gaze.revision`. Every other number is identical: `pass2-ner` leaked 20,727 B, FP 28,045 B; `rule-floor-extended` leaked 87,647 B, FP 5,433 B. Both trees clean (`dirty: false`). The later commits only add tests.

What the new column shows on main (`pass2-ner`):

| Label | Gold | Fail validator | Leaked B, valid gold | Leaked B, invalid gold |
| --- | ---: | ---: | ---: | ---: |
| CREDITCARDNUMBER | 126 | 96 | 0 | 1,400 |
| IBAN | 207 | 67 | 0 | 1,595 |
| TAXNUM | 212 | 210 | 28 | 1,853 |
| PHONENUMBER | 359 | 41 | 699 | 469 |
| EMAIL | 375 | 0 | 19 | 0 |

Every card and IBAN byte that leaks on main is checksum-invalid corpus gold, which the validators refuse on purpose.

**Tests:** `uv run --project scripts/bench python -m unittest discover -s scripts/bench` gives 419 OK (1 skipped), `render_benchmark_doc.py --check` is in sync, and the render-only unittest run passes. The new tests: an end-to-end `run_config` test showing invalid gold lands in the failed column and both columns add up to the unchanged label leak; bucket, absence and not-applicable tests; renderer mutation sweeps (every column moves the document, every source path moves the row); a column-order pin; a runner summary column test.

**Mutation probe (committed tree copy, failure set diffed against baseline): 9/9 RED.** Mutants: swapped buckets; not-applicable bucketed; split dropped from `run_config`; split on a not-applicable label; renderer swaps sources; renderer keeps non-applicable labels; table never emitted; malformed history accepted; runner summary columns swapped. The last one survived at first, and the runner summary test was added for it.

**Rust gates** (pinned 1.96.0): `cargo fmt --all -- --check` green, `cargo clippy --workspace --all-features --all-targets -- -D warnings` green, `cargo test --workspace --all-features`: see the comment below. No Rust file changed.

No visible UI change, so no screenshots. The doc table only appears once a new release row is appended.

## Structure review

1. Verdict: `implement` (done in scope).
2. Opportunity: one closed two-bucket partition (`GOLD_VALIDITY_BUCKETS` + `gold_validity_bucket`) shared by the scorer, the summary and the renderer, instead of repeated `applicable`/`validator_passed` conditionals.
3. Why: encodes the invariant that valid + invalid add up to the label leak and that not-applicable labels are never zero-filled.
4. Scope: bench scripts, tests, doc prose, CHANGELOG. Nothing in the product.
5. Validation: the identity diff, 419 unittest, 9/9 mutants, fmt and clippy.

## Five axes

1 Reliability: unchanged, and no validator was loosened. 2 Reversibility: n/a. 4 Trust: leaks now read by mechanism (validator veto on synthetic invalid values vs real misses). 5 Ergonomics: adopters can read the card/IBAN numbers correctly.

Refs todo 3648.

https://claude.ai/code/session_013qcdEYHZ72UpgEmCC2e7dZ


Naoray:
Rust gates on head 93eaf2ae under pinned 1.96.0: `cargo fmt --all -- --check` exit 0, `cargo clippy --workspace --all-features --all-targets -- -D warnings` exit 0, `cargo test --workspace --all-features` exit 0.

Naoray:
Independent review (pr-drain, brief 9964). Verdict: **CLEAN**.
- Scorecard identity, re-diffed with my own script on the author's files (sha256 569a1c23… base 9f1cd524, 69f6b0bd… head ab870b79, both `dirty:false`): 54 added fields, all `production_recall_by_gold_validity`; nothing removed; the only changed values are latency timings and `gaze.revision`.
- Split invariant on real data: across all 10 applicable validator blocks, valid + invalid leaked bytes equal the label's leaked bytes exactly (`RecallAccumulator.add` counts per span without merging, so this holds by construction).
- `python -m unittest discover -s scripts/bench`: 419 OK (1 skipped). `render_benchmark_doc.py --check`: in sync. `release-history.json` is unchanged, the v0.14.0 row carries no `validator_recall`, and `validator_recall_from_run` on the committed v0.14.0 scorecard's default arm returns None, so the v0.14.0 doc stays byte-identical. The doc diff vs main is prose additions only.
- My own mutants (not the author's list): (A) every applicable gold span counted as valid → `test_production_leak_is_split_by_gold_validity_without_moving_headline` FAILED; (B) renderer reads both leak columns from valid gold → 3 renderer tests FAILED.
- Structure: skip, opportunity none beyond the shared two-bucket partition this already adds.

